PR #14, opened 2025-10-21
Try to add microsoftInternal__NAV-219519

haoranpb:
Candidate: https://dynamicssmb2.visualstudio.com/Dynamics%20SMB/_git/46eb6939-f5e9-495b-b6e9-976998345260/pullrequest/219519

Try to add the PR to dataset again, suspect the previous failure was due to two Shopify PRs in the same version, so moved `microsoftInternal__NAV-211710` to `v26.4` instead.


haoranpb:
The build passed but the PR is still not usable: tests passed without the gold patch

PR #18, opened 2025-10-22
Try to add microsoftInternal__NAV-139511

haoranpb:
Candidate: https://dynamicssmb2.visualstudio.com/Dynamics%20SMB/_git/NAV/commit/ed5c415b4f2155c18718998a5f71a526c6b878dd?refName=refs/heads/master&path=/App/Apps/W1/SAF-T/test/src/SAFTTests.Codeunit.al


haoranpb:
Tests passed without the fix, probably was designed to fail in NO

PR #19, opened 2025-10-22
Only include Test Lib for container spinning up

haoranpb:
We don't need the standard tests, test lib only should be enough, this will speed things up for env setup


haoranpb:
running workflow to verify: https://github.com/microsoft/BC-Bench/actions/runs/18710718430

PR #23, opened 2025-10-23
Add support for BaseApp Datasets

haoranpb:
Adding support for Base Application dataset:
* Can ignore application property
* Collect dataset from path (W1 layers)


haoranpb:
The dataset failed verification because the code mismatch between master and releases/27.0, which is expected.

Will re-purpose the PR to add support for BaseApp instead

PR #32, opened 2025-10-24
try to add microsoftInternal__NAV-224832

haoranpb:


haoranpb:
Tests passed with the fix

PR #36, opened 2025-10-24
try to add microsoftInternal__NAV-223538

haoranpb:


haoranpb:
Tests won't pass, most likely due to the environment differences between NAV and AL-GO

PR #40, opened 2025-10-24
try to add microsoftInternal__NAV-221197

haoranpb:


haoranpb:
Tests won't pass, most likely due to the differences in env between NAV and AL-GO

PR #47, opened 2025-10-28
try to add microsoftInternal__NAV-219329

haoranpb:


haoranpb:
Tests passed before the fix

PR #52, opened 2025-10-28
try to add microsoftInternal__NAV-217395

haoranpb:


haoranpb:
Tests failed after fix, most likely due to environment differences

PR #56, opened 2025-10-29
try to add microsoftInternal__NAV-213820

haoranpb:
- [ ] problem statement must be updated due to screenshots


haoranpb:
Tests failed after the fix, most likely due to environment differences

PR #57, opened 2025-10-29
try to add microsoftInternal__NAV-215785

haoranpb:
- [ ] problem statement needs to be updated due screenshots


haoranpb:
Tests passed before the fix

PR #62, opened 2025-11-03
Try different model for GitHub Copilot

haoranpb:
Adding mode as a parameter so we can easily run different models and compare


haoranpb:
Too lazy to resolve conflicts, let's restart from main

PR #105, opened 2025-11-16
Revert "microsoftInternal__NAV-208320"

haoranpb:
Reverts microsoft/BC-Bench#98

The patches should filter on W1


haoranpb:
@Jiawen-CS FYI, we need to `uv run bcbench collect nav 208320--diff-path .\\App\\Layers\\W1\\`

So we don't include diffs from localizations

Jiawen-CS:
OK I will revert it.

PR #34, opened 2025-10-24
try to add microsoftInternal__NAV-224367

haoranpb:

PR #179, opened 2025-12-01
microsoftInternal__NAV-224367

Jiawen-CS:


haoranpb:
@Jiawen-CS realized that I tried this one before: #34

Might be a good idea to do a quick search before you try one, especially when there are few screenshots (I probably tried before..)

Jiawen-CS:
> @Jiawen-CS realized that I tried this one before: #34
> 
> Might be a good idea to do a quick search before you try one, especially when there are few screenshots (I probably tried before..)

Good idea.

PR #180, opened 2025-12-01
microsoftInternal__NAV-223313

Jiawen-CS:


haoranpb:
@Jiawen-CS , we haven't released `27.2` yet, so it won't work here. `27.0` will probably work.

Also, separate those two

Jiawen-CS:
> @Jiawen-CS , we haven't released `27.2` yet, so it won't work here. `27.0` will probably work.
> 
> Also, separate those two

OK, I will try this later.

PR #184, opened 2025-12-02
Add microsoftInternal__NAV-222846

Jiawen-CS:


haoranpb:
@Jiawen-CS See the changes under `GenJnlPostBatch`, the changes are adding an error. And tests are verifying the error massage => test has a dependency on the fix

Jiawen-CS:
> @Jiawen-CS See the changes under `GenJnlPostBatch`, the changes are adding an error. And tests are verifying the error massage => test has a dependency on the fix

Because there defined two 'PostingDateErr' strings, so we don't consider this PR?

PR #210, opened 2025-12-05
Cache artifacts instead of images

haoranpb:
Tried #209 , it didn't offer the expected speed up even when cache hit:
* With cache hit: https://github.com/microsoft/BC-Bench/actions/runs/19968738621/job/57267592970
  * 1 min to download
  * 5 mins to load
* Without cache hit: https://github.com/microsoft/BC-Bench/actions/runs/19964830453/job/57253819490
  * 6 mins to pull

Removing the cache for the image (will consider not using fresh machine instead), but cache the artifacts


haoranpb:
closed in favor of https://github.com/microsoft/BC-Bench/commit/a7c5484a349e998434043f6495951e5e78559cbf

PR #236, opened 2025-12-08
test things missing in new image

haoranpb:


haoranpb:
Got everything we need, we can setup node ourselves

PR #249, opened 2025-12-09
Add microsoft__BCApps-4392

Jiawen-CS:


haoranpb:
@Jiawen-CS Reading the bug itself, I don't there is enough context in the bug for the agent to solve it...

This is one of the things with bug fixes for new features. We usually don't add many context in the bug

Jiawen-CS:
> @Jiawen-CS Reading the bug itself, I don't there is enough context in the bug for the agent to solve it...
> 
> This is one of the things with bug fixes for new features. We usually don't add many context in the bug

Should we close it or just add more details by ourselves?

haoranpb:
> > @Jiawen-CS Reading the bug itself, I don't there is enough context in the bug for the agent to solve it...
> > This is one of the things with bug fixes for new features. We usually don't add many context in the bug
> 
> Should we close it or just add more details by ourselves?

let's close it. It is hard (impossible) to re-create the exact same context

Jiawen-CS:
Ok, now close.

PR #250, opened 2025-12-09
Add microsoft__BCApps-4384

Jiawen-CS:


haoranpb:
@Jiawen-CS the tests have dependency on the fix (it was more of a feature than bug fix tbh)

Jiawen-CS:
> @Jiawen-CS the tests have dependency on the fix (it was more of a feature than bug fix tbh)

OK, I got it now. Let me look another PRs before or after it.

PR #272, opened 2025-12-12
Save combine result in one single jsonl for easy download and analysis

haoranpb:


haoranpb:
@copilot this now re-upload all the files as well.

rename `combined_results` to `evaluation_summary`, so we could filter upload files with name `evaluation_summary.*`

PR #300, opened 2025-12-23
add microsoftInternal__NAV-215794

Jiawen-CS:


Jiawen-CS:
The namespace is "Microsoft.Bank.Deposit". Right here I used "Finance" as the area. Not sure whether is right.

haoranpb:
Unfortunately, this one's test has dependency on the fix, note the string

PR #312, opened 2025-12-29
Add microsoftInternal__NAV-213941

Jiawen-CS:


haoranpb:
Don't think the bug contains all the information needed to generate the fix, see internal bug in ADO 559286.

It linked to another bug fixed very similar issues, I'd assume the fix is heavily influenced by the previous fix

PR #329, opened 2026-01-02
Add microsoft__BCApps-4628

Jiawen-CS:


Jiawen-CS:
@haoranpb If you have time, can you check this one?

haoranpb:
@Jiawen-CS This we can't use due to dependency from test => fix, note the string `AutoCreateUnknownItemsDisabledErr`.

I think it's acceptable that we focus on `NAV` repo for now, we will try to achieve a diverse dataset by e.g. collecting from different namespaces

PR #354, opened 2026-01-06
Add microsoftInternal__NAV-207369

Jiawen-CS:


haoranpb:
There is a dependency from test => fix, see the label `ChangeExtendedTextErr`

PR #356, opened 2026-01-06
Add microsoftInternal__NAV-189991

Jiawen-CS:


Jiawen-CS:
For App/Apps/W1. maybe because NAV and Github repo would add features at the same period so sometimes it is hard to find correct version. Now I'm not going to close it now, and I will try to find the suitable base commit for it.

Jiawen-CS:
Same error happens a lot.

PR #392, opened 2026-01-12
Add microsoftInternal__NAV-185451

Jiawen-CS:


haoranpb:
Test passed before the fix

PR #412, opened 2026-01-16
Add microsoftInternal__NAV-191794

Jiawen-CS:


haoranpb:
The bug is linked to another one, doesn't contain all the context for it to be fixed.

PR #448, opened 2026-01-20
Add microsoftInternal__NAV-184494

Jiawen-CS:


haoranpb:
Tests have dependency on the fix, note the newly introduced fields

PR #449, opened 2026-01-20
Add microsoftInternal__NAV-184131

Jiawen-CS:
Use base commit from April.


Jiawen-CS:

PR #459, opened 2026-01-20
Add microsoftInternal__NAV-185450

Jiawen-CS:


haoranpb:
corrupt images in the bugs description. it also has few linked issues

PR #462, opened 2026-01-20
Add microsoftInternal__NAV-181052

Jiawen-CS:


haoranpb:
Repro steps in attachments

PR #557, opened 2026-02-02
Create SKILL.md for copilot

Jiawen-CS:
This pull request adds support for "skills" in the Copilot agent workflow, allowing repository-specific skill instructions to be discovered and used during agent runs. 

These changes enable the agent to discover and use repository-specific skills.

### DIFF
--- a/src/bcbench/agent/copilot/agent.py
+++ b/src/bcbench/agent/copilot/agent.py
@@ -13,7 +13,7 @@ from bcbench.config import get_config
 from bcbench.dataset import DatasetEntry
 from bcbench.exceptions import AgentError, AgentTimeoutError
 from bcbench.logger import get_logger
-from bcbench.operations import setup_custom_agent, setup_instructions_from_config
+from bcbench.operations import setup_custom_agent, setup_instructions_from_config, setup_copilot_skills
 from bcbench.types import AgentMetrics, EvaluationCategory, ExperimentConfiguration
 
 logger = get_logger(__name__)
@@ -34,6 +34,7 @@ def run_copilot_agent(entry: DatasetEntry, model: str, category: EvaluationCateg
     prompt: str = build_prompt(entry, repo_path, copilot_config, category, al_mcp=al_mcp)
     mcp_config_json, mcp_server_names = build_mcp_config(copilot_config, entry, repo_path, al_mcp=al_mcp)
     instructions_enabled: bool = setup_instructions_from_config(copilot_config, entry, repo_path)
+    copilot_skills: list[str] | None = setup_copilot_skills(copilot_config, entry, repo_path)
     custom_agent: str | None = setup_custom_agent(copilot_config, entry, repo_path)
     config = ExperimentConfiguration(mcp_servers=mcp_server_names, custom_instructions=instructions_enabled, custom_agent=custom_agent)
 
--- a/src/bcbench/agent/copilot/agent.py
+++ b/src/bcbench/agent/copilot/agent.py
@@ -13,7 +13,7 @@ from bcbench.config import get_config
 from bcbench.dataset import DatasetEntry
 from bcbench.exceptions import AgentError, AgentTimeoutError
 from bcbench.logger import get_logger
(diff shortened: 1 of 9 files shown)


src/bcbench/agent/copilot/agent.py
```diff
@@ -34,6 +34,7 @@ def run_copilot_agent(entry: DatasetEntry, model: str, category: EvaluationCateg
     prompt: str = build_prompt(entry, repo_path, copilot_config, category, al_mcp=al_mcp)
     mcp_config_json, mcp_server_names = build_mcp_config(copilot_config, entry, repo_path, al_mcp=al_mcp)
     instructions_enabled: bool = setup_instructions_from_config(copilot_config, entry, repo_path)
+    setup_copilot_skills(copilot_config, entry, repo_path)  # Copies skills to .github/skills/, CLI auto-discovers
     custom_agent: str | None = setup_custom_agent(copilot_config, entry, repo_path)
     config = ExperimentConfiguration(mcp_servers=mcp_server_names, custom_instructions=instructions_enabled, custom_agent=custom_agent)
```
haoranpb: You will have to add an extra field for `ExperimentConfiguration` as well, similar to `custom_instructions`

With the current implementation, we can do a run, and won't be able to tell if SKILLs are enabled in the final result

src/bcbench/operations/skills_operations.py
```diff
@@ -0,0 +1,39 @@
...
+
+def setup_copilot_skills(copilot_config: dict, entry: DatasetEntry, repo_path: Path) -> list[str] | None:
+    """
+    Setup skills in the repository if available.
+
+    Returns:
+        List of skill directory paths if skills are enabled and exist, None otherwise.
+    """
+    skills_config: dict = copilot_config["skills"]
+    skills_enabled: bool = skills_config["enabled"]
```
haoranpb: I'd simply combine them into one line, since we don't need the `skills_config`

src/bcbench/operations/skills_operations.py
```diff
@@ -0,0 +1,39 @@
...
+
+    if skills_enabled:
+        source_skills: Path = _get_source_instructions_path(entry.repo)
+        source_skills_dir = source_skills / "skills"
+
+        # Skip if skills folder doesn't exist for this repo
+        if not source_skills_dir.exists():
+            logger.info(f"No skills folder found at {source_skills_dir}, skipping")
+            return None
```
haoranpb: `raise` an error here instead.

If someone enabled skills, then the skills folder is probably not empty

src/bcbench/operations/skills_operations.py
```diff
@@ -0,0 +1,39 @@
...
+
+        github_dir: Path = repo_path / ".github"
+        skills_dir = github_dir / "skills"
+        copytree(source_skills_dir, skills_dir, dirs_exist_ok=True)
+
+        logger.info(f"Skills copied from {source_skills_dir} to {skills_dir}")
+        return [str(skills_dir)]
+
+    return None
```
haoranpb: you will need to return a `bool` for the ExperimentConfiguration mentioned above

haoranpb (changes requested):

haoranpb (approved):
Looks good, can you add some tests for this?

Jiawen-CS:
> Looks good, can you add some tests for this?

Now add tests for agent skills

src/bcbench/operations/skills_operations.py
```diff
@@ -0,0 +1,42 @@
...
+        # Remove existing skills directory to ensure clean state
+        if skills_dir.exists():
+            rmtree(skills_dir)
+
+        copytree(source_skills_dir, skills_dir)
+
+        logger.info(f"Skills copied from {source_skills_dir} to {skills_dir}")
+        return True
+
+    return False
```
haoranpb: Perhaps clearer?

```suggestion
    return skills_enabled
```

haoranpb (approved):

PR #559, opened 2026-02-03
Claude Code Agent Integration & Add Agent-Type

Jiawen-CS:
Related link: (https://github.com/microsoft/BC-Bench/issues/287)

This pull request adds support for Claude Code agent experiments in BC-Bench, aligning its configuration and setup with Copilot. It introduces agent-type-aware logic throughout the codebase, allowing shared configuration and instruction handling for both Claude and Copilot. The GitHub Actions workflow is updated to support Claude-specific options and tools, and the documentation is revised to reflect these changes and provide clear instructions for running Claude Code experiments.

These changes make it straightforward to run experiments with Claude Code using the same flexible configuration and setup as Copilot, while ensuring each agent's requirements are handled correctly.


src/bcbench/agent/shared/config.yaml
```diff
@@ -53,20 +53,25 @@ prompt:
     {% endif %}
 
 # controls:
-# 1. whether to copy custom instructions (`src/bcbench/agent/copilot/instructions/<sanitized-repo>/`) into the repo
+# 1. whether to copy custom instructions from `src/bcbench/agent/shared/instructions/<sanitized-repo>/`
+#    - Copilot: copies to repo/.github/ (CLI auto-reads)
+#    - Claude: copies to repo/.claude/ (passed via --append-system-prompt-file)
```
haoranpb: with the current implementation, this is not the case. We are not passing anything using `--append-system-prompt-file`.

Test locally and see if claude CLI will also auto-read, it's easier if they both work the same way

src/bcbench/agent/shared/config.yaml
```diff
@@ -53,20 +53,25 @@ prompt:
...
 # 2. whether to pass --disable-custom-instructions to copilot
 instructions:
   enabled: false
 
 # controls:
-# 1. whether to copy skills (`src/bcbench/agent/copilot/instructions/<sanitized-repo>/skills/`) into repo/.github/skills/
-# 2. CLI auto-discovers from repo/.github/skills/
+# 1. whether to copy skills from `src/bcbench/agent/shared/instructions/<sanitized-repo>/skills/`
+#    - Copilot: copies to repo/.github/skills/ (CLI auto-discovers)
```
haoranpb:
```suggestion
#    - Copilot: copies to repo/.github/skills/
```
Keep the two aligned

src/bcbench/operations/instruction_operations.py
```diff
@@ -8,8 +8,7 @@
 logger = get_logger(__name__)
 _config = get_config()
 
-
-def setup_instructions_from_config(copilot_config: dict, entry: DatasetEntry, repo_path: Path) -> bool:
+def setup_instructions_from_config(copilot_config: dict, entry: DatasetEntry, repo_path: Path, agent_type: str = "copilot") -> bool:
```
haoranpb: As discussed in person, the `agent_type` should be similar to `class EvaluationCategory(str, Enum)`.

This way, we define it once in `types.py` and we can re-use it everything

src/bcbench/operations/instruction_operations.py
```diff
@@ -59,11 +59,15 @@ def _get_source_instructions_path(repo_name: str) -> Path:
...
 
+    Instructions are stored in shared/instructions/ and used by both Copilot and Claude.
+
     Raises:
         FileNotFoundError: If instruction file doesn't exist
     """
     sanitized_name = repo_name.replace("/", "-")
-    instructions_path = _config.paths.copilot_dir / _config.file_patterns.copilot_instructions_dirname / sanitized_name
+    # Instructions are now in shared/instructions/ for all agents
+    shared_dir = _config.paths.copilot_dir.parent / "shared"
```
haoranpb: see `_config.paths.agent_share_dir`

src/bcbench/operations/skills_operations.py
```diff
@@ -24,8 +24,9 @@ def setup_agent_skills(agent_config: dict, entry: DatasetEntry, repo_path: Path)
         if not source_skills_dir.exists():
             raise FileNotFoundError(f"Skills folder not found for repository: {entry.repo} at {source_skills_dir}")
 
-        github_dir: Path = repo_path / ".github"
-        skills_dir = github_dir / "skills"
+        # Copilot reads from .github automatically, Claude uses .claude with explicit flag
+        target_dir: Path = repo_path / (".github" if agent_type == "copilot" else f".{agent_type}")
```
haoranpb: you can probably have a function here instead `target_dir: Path = get_target_dir(agent_type)`.

And re-use it in instruction operations

EXPERIMENTS.md
```diff
@@ -34,6 +34,21 @@ Steps for an experiment:
...
+Claude Code uses the same shared configuration as Copilot:
+
+Steps for an experiment:
+1. Create a new branch: `git checkout -b experiment/<meaningful-name>`
+2. Edit `src/bcbench/agent/shared/config.yaml` and optionally modify instruction markdown under `src/bcbench/agent/shared/instructions/<sanitized-repo>/` (see below)
+3. Locally run one entry: `uv run bcbench run claude <entry_id>` to ensure everything is setup correctly
+4. Create a draft PR (the default template will let you switch to the experiment template)
+5. In GitHub Actions: run workflow `claude-evaluation` after selecting your branch & model
+6. Start with **Test Run** (2 entries) → verify the changes are picked up in logs
+7. Run full evaluation
```
haoranpb: keep this short, just say it's the same as Copilot, just different workflow to trigger

haoranpb (changes requested):

src/bcbench/operations/instruction_operations.py
```diff
@@ -26,18 +27,19 @@ def setup_instructions_from_config(copilot_config: dict, entry: DatasetEntry, re
 
     if instructions_enabled:
         source_instructions: Path = _get_source_instructions_path(entry.repo)
-        github_dir: Path = repo_path / ".github"
+        # Copilot reads from .github automatically, Claude uses .claude with explicit flag
```
haoranpb: Don't claude also read from `.claude` automatically?

src/bcbench/operations/instruction_operations.py
```diff
@@ -4,19 +4,20 @@
...
 from bcbench.dataset import DatasetEntry
 from bcbench.logger import get_logger
+from bcbench.types import AgentType
 
 logger = get_logger(__name__)
 _config = get_config()
 
-
-def setup_instructions_from_config(copilot_config: dict, entry: DatasetEntry, repo_path: Path) -> bool:
+def setup_instructions_from_config(copilot_config: dict, entry: DatasetEntry, repo_path: Path, agent_type: AgentType = AgentType.COPILOT) -> bool:
```
haoranpb: Let's not give it default values, so it will have to be explictly set

src/bcbench/operations/instruction_operations.py
```diff
@@ -59,11 +61,15 @@ def _get_source_instructions_path(repo_name: str) -> Path:
...
 
+    Instructions are stored in shared/instructions/ and used by both Copilot and Claude.
+
     Raises:
         FileNotFoundError: If instruction file doesn't exist
     """
     sanitized_name = repo_name.replace("/", "-")
-    instructions_path = _config.paths.copilot_dir / _config.file_patterns.copilot_instructions_dirname / sanitized_name
+    # Instructions are now in shared/instructions/ for all agents
+    shared_dir = _config.paths.agent_share_dir
```
haoranpb: `shared_dir` variable here is not necessary

src/bcbench/operations/instruction_operations.py
haoranpb: same everywhere else

src/bcbench/types.py
```diff
@@ -71,6 +71,17 @@ def is_empty(self) -> bool:
...
 
+class AgentType(str, Enum):
+    COPILOT = "copilot"
+    CLAUDE = "claude"
+
+    def get_target_dir(self, repo_path: Path) -> Path:
+        """Get target directory for agent-specific files based on agent type."""
+        if self == AgentType.COPILOT:
+            return repo_path / ".github"
+        return repo_path / f".{self}"
```
haoranpb: let's be explicity and have a case statement instead, something like blow

```python
case self:
    AgentType.Copilot:
        return repo_path / ".github"
    AgentType.Claude:
        pass
    _:
        raise "Not valida AgentType"
```

And you can do a test like `test_category_exhaustiveness.py`, so people don't forget to add new agenttype to the case statement

haoranpb (changes requested):
One more thing, in order to support al mcp server, you will have to change the `claude-evaluation.yml`, use `copilot-evaluation.yml` as reference.

Maybe also adjust the title and description of the PR, claude code is already there for Copilot, but it's imcomplete => only supports the baseline runs for bug-fixing. Now you are making it support Test-Generation and all the experiemnts.

Finally, don't forget to manual test locally.

src/bcbench/operations/instruction_operations.py
Jiawen-CS: Directly use _config.paths.agent_share_dir now.

src/bcbench/operations/instruction_operations.py
Jiawen-CS: Wrong comment. Delete it now.

src/bcbench/operations/instruction_operations.py
Jiawen-CS: move " = AgentType.COPILOT"

src/bcbench/agent/shared/mcp.py
```diff
@@ -49,23 +50,23 @@ def _build_server_entry(server: dict[str, Any], template_context: dict[str, Any]
...
             args: list[str] = server["args"]
             rendered_args = [Template(arg).render(**template_context) for arg in args]
             return server_name, {
-                "type": server_type,
+                "type": server_type if agent_type == AgentType.COPILOT else "stdio", 
                 "command": server["command"],
                 "args": rendered_args,
-                "tools": tools,
+                "tools": tools if agent_type == AgentType.COPILOT else None,
             }
```
haoranpb, line 59: I assume these if statements are needed because copilot and claude share different format for mcp configuration. The tricky part here: if we were to introduce a new coding agent (e.g. codex), the created mcp configuration might not work.

Create functions like `get_mcp_tools` and `get_mcp_server_type` instead, and inside, use case statement to explicitly try to each AgentType, and raise an exception if not matched. This way, if we do introduce a new coding agent, the logic here will have to be explicitly implemented.

haoranpb (changes requested):

src/bcbench/agent/shared/mcp.py
Jiawen-CS: That's a good way. Will try that. Thanks

haoranpb (approved):
This looks pretty good now, resolve the conflicts we should be good to go.

One minor thing: consider merging the `test_agent_type_exhaustiveness` and `test_category_exhaustiveness`

Jiawen-CS:
@haoranpb Could you please `uv run bcbench run claude` locally to check again? Because I don't have access to it since this month. If everything is OK, then we can close the PR.

PR #584, opened 2026-03-08
update copilot and refresh models

haoranpb:
Getting the latest `gpt-5.4` and also removing older models that are scheduled to be deprecated.

Ref: https://github.blog/changelog/2026-03-02-upcoming-deprecation-of-gemini-3-pro-and-gpt-5-1-models/


haoranpb:
@Copilot remove the disable parallel tool execution flag for copilot, the latest version doesn't support it.

PR #591, opened 2026-03-18
add CFE dataset

Jiawen-CS:
This project investigates transfer failures in large language models (LLMs) when generating code for niche programming languages, using AL as a case study.

We design BC-Bench-CF, a benchmark suite that includes realistic AL development tasks and minimal counterfactual variants. The goal is to evaluate not only functional correctness, but also robustness to small specification changes and sensitivity to AL-specific execution semantics.

Our analysis is grounded in a layered failure framework, which attributes model errors to different abstraction levels, including syntax, validation semantics, event-driven paradigms, workflow composition, and ecosystem constraints.


haoranpb:
You can mark the PR as draft, so it won't get accidentally merged

Jiawen-CS:
Notice that we have big changes in the main branch. To keep alignment with it, we now close this PR and would create a new branch for master thesis.

PR #593, opened 2026-03-18
Refactor dataset model for extensibility across evaluation categories

haoranpb:
Introduces a base/subclass hierarchy for dataset entries so new evaluation categories with different schemas and evaluation logic can be added without modifying existing code.


src/bcbench/operations/setup_operations.py
```diff
@@ -51,17 +51,23 @@ def setup_repo_prebuild(entry: DatasetEntry, repo_path: Path) -> None:
     checkout_commit(repo_path, entry.base_commit)
 
 
-def setup_repo_postbuild(entry: DatasetEntry, repo_path: Path, category: EvaluationCategory) -> None:
-    """Setup repository after building - apply patches and copy problem statements.
```
haoranpb, line 55: Might need to rethink these. Some setups are category specific

Jiawen-CS (approved):

PR #610, opened 2026-04-12
Introducing Code Review category

haoranpb:
A POC showing how Code Review category can be implemented.

Try it out locally:
```shell
# See the placeholder task in the dataset
uv run bcbench dataset view synthetic__performance-001 --category code-review

# See the default copilot review live
uv run bcbench -v evaluate copilot synthetic__performance-001 --category code-review --repo-path C:\depot\BCApps
```

[Getting Started](https://github.com/microsoft/BC-Bench/blob/main/CONTRIBUTING.md)

## Code Review team to implement

- [x] Replace the dataset, see `codereview.jsonl`
- [x] Implement the scoring and evaluation metrics calculation, we currently have placeholder metrics: count of comments
- [x] Add the AL Code Review skill 

~Leaderboard related logic is intentionally left unimplemented for now.~


tools/code-review/run_all_evals.ps1
haoranpb, line 1: dedicated scripts are designed to be stored under `tools/<category>`

src/bcbench/agent/copilot/agent.py
```diff
@@ -31,6 +31,11 @@ def run_copilot_agent(
     config_file = Path(__file__).parent.parent / "shared" / "config.yaml"
     copilot_config = yaml.safe_load(config_file.read_text())
 
+    is_code_review_bcapps = category == EvaluationCategory.CODE_REVIEW and entry.repo == "microsoft/BCApps"
+    if is_code_review_bcapps:
+        copilot_config.setdefault("instructions", {})["enabled"] = True
+        copilot_config.setdefault("skills", {})["enabled"] = True
+
```
haoranpb: Interesting work around here.

The designed behavior is the default scenario should always be the **baseline**, so in your case, the default Copilot Review Agent (i.e. /review).

And an experiment branch has to be created to enable custom configurations (e.g. different skills)

src/bcbench/agent/shared/config.yaml
```diff
@@ -54,6 +54,16 @@ prompt:
     {{task}}
     {% endif %}
 
+  code-review-template: |
+    @al-code-review
```
haoranpb: The default template should point to the out-of-the-box Copilot Review Agent (note, slash command is for activating skills)

```suggestion
    /review
```

In order to experiemnt with the new al-code-review skill, a experiment branch will have to be created and update this.

src/bcbench/evaluate/codereview.py
```diff
@@ -0,0 +1,63 @@
...
+
+    Code review does not require a BC container — the agent reviews a patch
+    and produces review comments without building or running tests.
+    """
+
+    def setup_workspace(self, entry: CodeReviewEntry, repo_path: Path) -> None:
+        """Setup workspace for code review.
+        
+        For code-review, we only clean/reset the repo. The patch is passed to
+        the agent for review but not applied to the working directory.
```
haoranpb: I think this is an excellent place to run experiments.

Remember you mentioned some folks mentioned that providing the agent with full repository will get better result. Probably interesting to try that here.

I would assume the default scenior is with diff/patch applied, or at least this is what `/review` does

> ... Analyzes staged/unstaged changes and branch diffs ...

src/bcbench/evaluate/codereview.py
```diff
@@ -0,0 +1,63 @@
...
+            output = review_output_file.read_text(encoding="utf-8")
+        else:
+            logger.error(f"No review generated for {context.entry.instance_id}")
+            raise RuntimeError(f"No review generated for {context.entry.instance_id}")
+
+        result = CodeReviewResult.create_success(
+            context,
+            output=output,
+            expected_comments=context.entry.expected_comments,
+            line_tolerance=context.entry.match_line_tolerance,
```
haoranpb: I see you have majorify of the metrics calculation logic in the `CodeReviewResult` class, move them here instead.

The result class itself should not contain much logic, but try to stay immutable.

src/bcbench/results/codereview.py
```diff
@@ -0,0 +1,284 @@
...
+    @model_validator(mode="after")
+    def _parse_comments_from_output(self) -> Self:
+        if not self.generated_comments and self.output:
+            parsed_comments, valid_output = _parse_review_output(self.output)
+            object.__setattr__(self, "generated_comments", parsed_comments)
+            object.__setattr__(self, "valid_review_output", valid_output)
+        elif not self.output.strip():
+            object.__setattr__(self, "valid_review_output", False)
+
+        matches = _match_comments(self.expected_comments, self.generated_comments, self.line_tolerance)
```
haoranpb: See comments in `evaluate/codereview.py`, move the comments parsing and metrics calculation in the evaluate pipeline as much as possible.

We shouldn't need those `__setattr__` if the metrics are already calculated before model validation.

src/bcbench/results/summary.py
```diff
@@ -165,6 +165,78 @@ def from_results(cls, results: Sequence[BaseEvaluationResult], run_id: str) -> "
...
+            "generated_comment_count": self.generated_comment_count,
+            "expected_comment_count": self.expected_comment_count,
+            "matched_comment_count": self.matched_comment_count,
+            "incorrect_comment_count": self.incorrect_comment_count,
+            "missed_comment_count": self.missed_comment_count,
+            "precision": round(self.precision * 100, 1),
+            "recall": round(self.recall * 100, 1),
+            "f1": round(self.f1 * 100, 1),
+            "severity_mae": round(self.severity_mae, 3),
+            "valid_review_output_rate": round(self.valid_review_output_rate * 100, 1),
```
haoranpb: For display summery, I'd trim the metrics aggresively, those `*comment_counts` are included in `precision` and `recall`

```suggestion
            "precision": round(self.precision * 100, 1),
            "recall": round(self.recall * 100, 1),
            "f1": round(self.f1 * 100, 1),
            "severity_mae": round(self.severity_mae, 3),
```

haoranpb: But anyhow, we didn't prepare the PR for the leaderboard yet anyway, skip for now

haoranpb (commented):
Solid progress.

One thing we should discuss: do we want this synthatic dataset? Or do we want to invest in real-world production PRs?

src/bcbench/results/codereview.py
```diff
@@ -0,0 +1,158 @@
...
+            "severity_mae": round(self.severity_mae, 3),
+            "valid_review_output": self.valid_review_output,
+        }
+
+    @property
+    def display_row(self) -> dict[str, str]:
+        return {
+            "Generated": str(len(self.generated_comments)),
+            "Matched": str(self.matched_comment_count),
+            "Expected": str(len(self.expected_comments)),
```
haoranpb, line 213: Probably not needed for the single run display, the results will be available in Kusto for analysis, here is just a quick summery

src/bcbench/results/codereview.py
```diff
@@ -0,0 +1,158 @@
...
+
+    return matched
+
+
+def _severity_mae(matched_pairs: list[tuple[ReviewComment, ReviewComment]]) -> float:
+    if not matched_pairs:
+        return 0.0
+    total_error: int = sum(abs(expected.severity.level - generated.severity.level) for expected, generated in matched_pairs)
+    return total_error / len(matched_pairs)
+
```
haoranpb: docstring for this function, and fullname instead of shorthand

src/bcbench/results/codereview.py
haoranpb: Acutally, maybe verse versa, we probably don't care about metrics like f1 score per task, it's the aggregate that really matters

src/bcbench/results/codereview.py
```diff
@@ -0,0 +1,228 @@
...
+        return start - line
+    return line - effective_end
+
+
+def _match_comments(
+    expected_comments: list[ReviewComment],
+    generated_comments: list[ReviewComment],
+    line_tolerance: int,
+) -> list[tuple[ReviewComment, ReviewComment]]:
+    """Greedily pair each expected comment with the nearest unused generated comment in the same file."""
```
kmhansen: nit: better to do bipartite matching

src/bcbench/results/codereview.py
```diff
@@ -0,0 +1,228 @@
...
+            if distance > line_tolerance:
+                continue
+
+            if best_distance is None or distance < best_distance:
+                best_distance = distance
+                best_index = index
+
+        if best_index is not None:
+            used_generated.add(best_index)
+            matched.append((expected, generated_comments[best_index]))
```
kmhansen: Hmmh. So not consideration of whether comments match semantically?

Groenbech96 (commented):
I would consider using a LLM as a judge to compare the review comment. Probably instruct that is has to be the same intent with the comment.

src/bcbench/results/codereview.py
WaelAbuSeada: Yea, I had this in initially, but not sure if this is something we want in v1 to avoid going down the rabbit hole of judging the judge. 

I can simple LLM judge that do semantic comparison

WaelAbuSeada:
Take a look at the score aggregation

src/bcbench/results/codereview.py
WaelAbuSeada: I think both matters

src/bcbench/results/codereview.py
WaelAbuSeada: Good point. Sticking with greedy for v1 — expected comment lists per entry are tiny (1-5), so the worst-case mis-pairing has at most low single-digit impact on the score, and the implementation stays straightforward. Happy to swap in a Hungarian/min-cost assignment if we see actual cases where greedy mis-pairs in a meaningful way.

src/bcbench/dataset/codereview.py
```diff
@@ -0,0 +1,79 @@
...
+        loc = f"{self.file}:{self.line_start}"
+        if self.line_end and self.line_end != self.line_start:
+            loc += f"-{self.line_end}"
+        return f"[{self.severity}] {loc}: {self.body}"
+
+
+class CodeReviewEntry(BaseDatasetEntry):
+    """Dataset entry for the code-review category."""
+
+    domain: str | None = None
```
haoranpb: This property looks unused, I see it's already present on the comments

src/bcbench/evaluate/codereview.py
```diff
@@ -0,0 +1,155 @@
...
+
+        if line.startswith("+"):
+            current_content.append(line[1:])
+            continue
+
+        if line.startswith("\\ No newline at end of file"):
+            continue
+
+    flush_current()
+    return materialized_paths
```
haoranpb: The code here is unnecessary, it is trying to handle incorrect git patch, which shouldn't happen in the code review production envrionment, where patch is collected by `git diff`.

If a patch won't apply, we should fix the patch instead, see https://github.com/microsoft/BC-Bench/pull/610/changes/c9193e5d464d4ac0e7bdd0a9331dadd0158b0586

src/bcbench/evaluate/codereview.py
```diff
@@ -0,0 +1,155 @@
...
+                logger.info(f"Materialized {len(materialized_paths)} file(s) from simplified review patch for {entry.instance_id}")
+
+            if paths := _patched_paths(entry.patch):
+                subprocess.run(
+                    ["git", "add", "-N", "--", *paths],
+                    cwd=repo_path,
+                    stdout=subprocess.DEVNULL,
+                    stderr=subprocess.PIPE,
+                    check=True,
+                )
```
haoranpb: Same here, if `apply_patch` fails, we should fix the patch instead, so it's aligned with production environment

src/bcbench/evaluate/codereview_judge.py
```diff
@@ -0,0 +1,117 @@
...
+
+from bcbench.config import get_config
+from bcbench.dataset.codereview import ReviewComment
+
+_config = get_config()
+
+JUDGE_RESULT_FILE = "judge_results.json"
+
+_JUDGE_PROMPT_TEMPLATE = """\
+You are a code review evaluation judge. Your task is to determine whether pairs of code review \
```
haoranpb: Not sure if all those `\` escape is necessary here

src/bcbench/evaluate/codereview_judge.py
```diff
@@ -0,0 +1,117 @@
...
+    if not result_path.exists():
+        return [True] * num_pairs
+
+    try:
+        raw = json.loads(result_path.read_text(encoding="utf-8"))
+    except (json.JSONDecodeError, OSError):
+        return [True] * num_pairs
+
+    if not isinstance(raw, list):
+        return [True] * num_pairs
```
haoranpb: Shouldn't we `raise` or return `[False]` when the judge fails?

src/bcbench/evaluate/codereview_judge.py
```diff
@@ -0,0 +1,117 @@
...
+                f"--model={model}",
+                f"--prompt={prompt.replace(chr(10), ' ')}",
+            ],
+            cwd=str(work_dir),
+            capture_output=True,
+            timeout=_config.timeout.agent_execution,
+            check=True,
+        )
+    except (subprocess.TimeoutExpired, subprocess.CalledProcessError, OSError):
+        return matched_pairs
```
haoranpb: same here, we are default matching everything if LLM judge fails

src/bcbench/evaluate/codereview.py
```diff
@@ -0,0 +1,155 @@
...
+            result = CodeReviewResult.create_invalid(context, output, context.entry.expected_comments)
+        else:
+            structural_matches = match_comments(
+                context.entry.expected_comments,
+                generated_comments,
+                context.entry.match_line_tolerance,
+            )
+            validated_matches = judge_comment_matches(
+                structural_matches,
+                model=context.model,
```
haoranpb: Here we are using the same model to judge its own output.

If we compare the results between `gpt-5.2` and `claude-sonnet-4.6`, we will be comparing two things:
1. the model's abilily to review AL code
2. the model's ability to match comments

The judge model should probably be fixed here, no matter the experiment model, maybe the LTS model: https://docs.github.com/en/copilot/concepts/models/fallback-and-lts-models

src/bcbench/operations/hooks_operations.py
```diff
@@ -13,31 +13,37 @@
...
+    # already works for bug-fix/test-gen runs on self-hosted runners.
     hooks_config = {
         "version": 1,
         "hooks": {
             "preToolUse": [
                 {
                     "type": "command",
-                    "powershell": f'powershell -ExecutionPolicy Bypass -File "{script_path}"',
+                    "bash": f'python3 "{py_script}"',
+                    "powershell": f'powershell -ExecutionPolicy Bypass -File "{ps_script}"',
```
haoranpb: It is probably enough to simply rename replace `"powershell"` to `"command"`, which is OS agnostic

src/bcbench/results/codereview.py
```diff
@@ -0,0 +1,433 @@
...
+    if not matched_pairs:
+        return 0.0
+    total_error: int = sum(abs(expected.severity.level - generated.severity.level) for expected, generated in matched_pairs)
+    return total_error / len(matched_pairs)
+
+
+class CodeReviewResult(BaseEvaluationResult):
+    """Result for the code-review category."""
+
+    domain: str = "unknown"
```
haoranpb: Should `domain` only reside in `ReviewComment`?

src/bcbench/results/codereview.py
```diff
@@ -0,0 +1,433 @@
...
+    table = Table(title=title, title_justify="left", title_style="bold cyan", show_header=True, header_style="bold")
+    for column in columns:
+        table.add_column(column, justify="right")
+    table.add_row(*row)
+    return table
+
+
+def _with_comment_domains(generated_comments: list[ReviewComment], domain: str) -> list[ReviewComment]:
+    """Stamp the entry domain onto comments that have no explicit domain. All comments are kept."""
+    return [comment if comment.domain else comment.model_copy(update={"domain": domain}) for comment in generated_comments]
```
haoranpb: Hmm, bit confused now what is the relationship between domain and the code review result class

src/bcbench/results/codereview.py
```diff
@@ -0,0 +1,433 @@
...
+            **cls._base_fields(context),
+            domain=domain,
+            output=output,
+            expected_comments=expected_comments,
+            generated_comments=generated_comments,
+            line_tolerance=line_tolerance,
+            valid_review_output=True,
+            matched_comment_count=matched_count,
+            incorrect_comment_count=max(0, len(generated_comments) - matched_count),
+            missed_comment_count=max(0, len(expected_comments) - matched_count),
```
haoranpb: Get rid of `max`? If the value gets under 0, we probably want to know what happened

src/bcbench/results/base.py
```diff
@@ -87,6 +97,15 @@ def display_row(self) -> dict[str, str]:
...
         return {}
 
+    @property
+    def sort_key(self) -> tuple[object, ...]:
+        """Sort key for the detailed results table.
+
+        Default orders by natural-sorted instance_id (so ``-002`` precedes ``-010``).
+        Subclasses can prepend category-specific keys (e.g. domain for code-review).
+        """
+        return (natural_sort_key(self.instance_id),)
```
haoranpb: Not sure this is necessary on the base class, could be specific on the code review sub class

haoranpb (commented):
We also need to add the display on the GitHub Page, similar to https://github.com/microsoft/BC-Bench/blob/main/docs/bug-fix.md, but that could be a follow up after we have some results to show

src/bcbench/operations/hooks_operations.py
haoranpb: I will take this one

src/bcbench/results/display.py
```diff
@@ -51,29 +48,27 @@ def create_console_summary(results: Sequence[BaseEvaluationResult], summary: Eva
...
     console.print(table)
     console.print()
 
 
-def _get_short_error_message(error_message: str | None) -> str:
-    """Extract the first line of an error message for summary display."""
-    if not error_message:
-        return ""
-    first_line = error_message.split("\n")[0].rstrip(":")
-    return first_line.replace("|", "\\|")
```
haoranpb, line 75: I think the error message column here is still necessary for all categories, a common one could be: copilot CLI timeout

.github/workflows/copilot-evaluation.yml
```diff
@@ -144,7 +145,9 @@ jobs:
         if: always()
         with:
           name: evaluation-results-${{ github.run_id }}-${{ matrix.entry }}
-          path: ${{ env.EVALUATION_RESULTS_DIR }}/**/*.jsonl
+          path: |
+            ${{ env.EVALUATION_RESULTS_DIR }}/**/*.jsonl
+            ${{ env.EVALUATION_RESULTS_DIR }}/**/*.log
```
haoranpb: the debug log might include too much information, what information are we trying to extrat here? Maybe a seperate log file

gggdttt: Good catch. The only thing we extract from these logs is the turn count, which is already parsed into the result .jsonl at eval time — so the raw *.log  was only useful for occasional manual debugging -  We can re-run locally with debug logs when we need deep debugging.

gggdttt: I shall remove the uploading of log

src/bcbench/dataset/codereview.py
gggdttt: Agree - it's duplicated with the metadata.area. Use that one.

src/bcbench/evaluate/codereview.py
gggdttt: Agreed. All 81 dataset patches are real output with proper hunk headers, so this fallback was unreachable in production

src/bcbench/evaluate/codereview.py
gggdttt: Agree - done

src/bcbench/evaluate/codereview.py
gggdttt: Support to change the model - default is gpt-5.3-codex but user can change it

src/bcbench/evaluate/codereview_judge.py
gggdttt: cleaned

src/bcbench/results/base.py
gggdttt: Good call. Moved the default ordering out of BaseEvaluationResult — the display layer display.py now owns the generic fallback So the framework base no longer carries a presentation/sort concern, and the code-review-specific behavior is fully in its subclass.

tools/code-review/audit_gold.py
```diff
@@ -0,0 +1,152 @@
+"""Intrinsic gold audit for code-review: validate each expected comment's location
```
gggdttt: This is used to eval the quality of our gold answers - but not sure if it should be merged in this PR or in future.

src/bcbench/results/codereview.py
gggdttt: Agreed. `matched_count` is bounded by both list lengths (bipartite matching), so these subtractions can never legitimately go negative — the max(0, …) was unreachable defensive clamping. Removed it; the `ge=0` field validators will now raise loudly if a real bug ever produced a negative.

src/bcbench/results/codereview.py
gggdttt: Agree - Removed

gggdttt:
> We also need to add the display on the GitHub Page, similar to https://github.com/microsoft/BC-Bench/blob/main/docs/bug-fix.md, but that could be a follow up after we have some results to show

https://dynamicssmb2.visualstudio.com/Dynamics%20SMB/_workitems/edit/639953 Added this deliverable

src/bcbench/results/codereview.py
gggdttt: Added and rename it to `_severity_mean_absolute_error`

src/bcbench/agent/shared/config.yaml
```diff
@@ -50,6 +50,20 @@ prompt:
     {{task}}
     {% endif %}
 
+  code-review-template: |
+    /review
+
+    Review the current branch changes and save findings to a file named "review.json" in the repository root.
+
+    Review ONLY the current working-tree AL file changes for this evaluation entry.
```
haoranpb: The goal of this prompt here is to be simple and align as much as possbile as your production prompt, maybe not `evaluation entry`

src/bcbench/evaluate/codereview.py
```diff
@@ -0,0 +1,96 @@
...
+
+        apply_patch(repo_path, entry.patch, f"{entry.instance_id} review patch")
+        if paths := _patched_paths(entry.patch):
+            subprocess.run(
+                ["git", "add", "-N", "--", *paths],
+                cwd=repo_path,
+                stdout=subprocess.DEVNULL,
+                stderr=subprocess.PIPE,
+                check=True,
+            )
```
haoranpb, line 46: don't think this is needed anymore, I believe it was added when the full file patches couldn't apply

src/bcbench/evaluate/codereview.py
```diff
@@ -0,0 +1,96 @@
...
+    Code review does not require a BC container. We materialize the dataset patch
+    as local git changes so the agent can review the branch diff directly.
+    """
+
+    def setup_workspace(self, entry: CodeReviewEntry, repo_path: Path) -> None:
+        """Setup workspace for code review by applying the entry patch as local changes."""
+        setup_repo_prebuild(entry, repo_path)
+        if not entry.patch.strip():
+            logger.warning(f"Entry {entry.instance_id} has empty patch; review will run on clean workspace")
+            return
```
haoranpb: this check is not needed, if the `patch` is empty, the `apply_patch` below will fail

src/bcbench/evaluate/codereview_judge.py
```diff
@@ -0,0 +1,147 @@
...
+
+    # A pair the judge never returned a verdict for counts as not confirmed.
+    return [results_by_pair.get(i + 1, False) for i in range(num_pairs)]
+
+
+def _find_copilot() -> str | None:
+    return shutil.which("copilot.exe") or shutil.which("copilot.cmd") or shutil.which("copilot")
+
+
+def judge_comment_matches(
```
haoranpb, line 92: More of a brainstorm, I wonder if there is a deterministic sementic check instead

src/bcbench/results/codereview.py
haoranpb: Still on the Result

src/bcbench/results/base.py
```diff
@@ -12,6 +13,15 @@
 logger = get_logger(__name__)
 
 
+def natural_sort_key(value: str) -> tuple[object, ...]:
```
haoranpb: Move this out to CodeReview

src/bcbench/results/display.py
```diff
@@ -5,15 +5,20 @@
...
 from bcbench.logger import get_logger
-from bcbench.results.base import BaseEvaluationResult
+from bcbench.results.base import BaseEvaluationResult, natural_sort_key
 from bcbench.results.summary import EvaluationResultSummary, calculate_average_tool_usage
 
 logger = get_logger(__name__)
 console = Console()
 
 
+def _detail_sort_key(result: BaseEvaluationResult) -> tuple[object, ...]:
```
haoranpb: Not sure if the changes in this file is due to merge conflict or something. I tried to make code generic in https://github.com/microsoft/BC-Bench/pull/664/changes

src/bcbench/agent/shared/config.yaml
gggdttt: Change to 
"Review only the uncommitted working-tree changes (git diff HEAD); do not compare commits such as HEAD~1..HEAD or origin/main.
Save the review findings to a file named "review.json" in the repository root (write an empty array if there are no findings).
"

src/bcbench/evaluate/codereview.py
gggdttt: agreed, removed it.

src/bcbench/evaluate/codereview.py
gggdttt: I'd keep this one. It's not about whether the patch applies — it's about diff visibility. All 81 code-review entries are new-file diffs, so after git apply the .al files are untracked. Verified in a scratch repo: git diff HEAD shows 0 lines for a new untracked file and 7 lines after git add -N. Since the agent is told to review git diff HEAD, dropping -N makes it see no changes and recall goes to zero. Added a short comment so the intent is clear.

src/bcbench/results/display.py
gggdttt: You are right - I remove it from this  PR,

haoranpb:
The only thing I have left:
1. https://github.com/microsoft/BC-Bench/pull/610#discussion_r3457611955
2. I think the latest run failed but the workflow is green is bit concerning. Two things could be hardened:
    a. The comment parsing should have failed, causing the whole workflow to fail or get an error message
    b. The `create_invalid` is bit too generic, we don't have visibility on what exactly failed. Consider do some more explicity error handling in `parse_review_output`.

Jiawen-CS (approved):

PR #615, opened 2026-04-16
Add counterfactual dataset

Jiawen-CS:
This project investigates transfer failures in large language models (LLMs) when generating code for niche programming languages, using AL as a case study.

We design BC-Bench-CF, a benchmark suite that includes realistic AL development tasks and minimal counterfactual variants. The goal is to evaluate not only functional correctness, but also robustness to small specification changes and sensitivity to AL-specific execution semantics.

Our analysis is grounded in a layered failure framework, which attributes model errors to different abstraction levels, including syntax, validation semantics, event-driven paradigms, workflow composition, and ecosystem constraints.


evaluator/counterfactual_scores.py
haoranpb, line 1: FYI, this is only needed if you are uploading data to braintrust

PR #637, opened 2026-05-01
Enforce additional ruff rules and bump tooling versions

haoranpb:
Started from removing two unused module-level globals that no linter caught. Turns out we were under-using ruff and ty (old versions, minimal rules).

One rule per commit, each justified.

## Added

- `TRY400` — `logger.error()` in except → `logger.exception()` (17 fixes)
- `LOG`, `G` — guardrails for logger misuse, no current violations
- `PT` — pytest patterns, 15 fixes (composite asserts, broad `pytest.raises`, fixture style)
- `ANN` — type annotations in `src/` and `evaluator/`, 32 fixes
- `T20` — caught 2 stale debug prints + 1 missing logger call
- `FURB`, `ISC`, `ICN`, `FA`, `PYI`, `SLOT`, `ASYNC` — zero-cost guardrail bundle
- `TID` (full), `TRY` (full minus 3) — promoted from sub-rules, caught 11 redundant `{e}` in `logger.exception()`
- Bumped ruff 0.13→0.15.12, ty 0.0.1a25→0.0.33 — caught 2× `class X(str, Enum)` → `StrEnum`

## Skipped (with reasons in commit history)

`S` (too noisy), `EM` (low value, lots of churn), `ERA` (too many false positives), `PERF` (controversial)

## Open follow-ups

- `ANN401` still ignored — `**kwargs: Any` in `results/base.py` for kwarg forwarding. Tried `object` and narrow unions, both fail; a `TypedDict` refactor works but verbose. Wants more thinking.
- ty pre-commit hook silently passes — `uvx`-spawned env can't import `bcbench`, hides errors involving our own types. Fix included in this PR (switched to `uv run` + `language: system`); 12 pre-existing type errors in `tests/test_evaluation_summary.py` now visible, to be fixed separately.

## Validation

`uv run ruff check .` clean · `uv run pytest tests/` 400 passed · `uv run pre-commit run --all-files` clean


haoranpb:
@copilot resolve the merge conflicts in this pull request

kmhansen (approved):

PR #656, opened 2026-05-29
Add claude-opus-4.8 model and bump version to 0.5.4

ventselartur:
Add claude-opus-4.8 to evaluation workflows and CLI options.

### DIFF
--- a/.github/workflows/claude-evaluation.yml
+++ b/.github/workflows/claude-evaluation.yml
@@ -106,7 +106,7 @@ jobs:
           echo "$HOME\.dotnet\tools" >> $env:GITHUB_PATH
 
       - name: Install Claude Code
-        run: npm install -g @anthropic-ai/claude-code@2.1.116
+        run: npm install -g @anthropic-ai/claude-code@2.1.160
 
       - name: Run Claude Code for entry ${{ matrix.entry }}
         timeout-minutes: 120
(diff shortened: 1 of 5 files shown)


.github/workflows/claude-evaluation.yml
```diff
@@ -14,6 +14,7 @@ on:
         options:
           - "claude-sonnet-4-6"
           - "claude-opus-4-7"
```
haoranpb: for claude code, they usually drop the old model when introduce the one

src/bcbench/cli_options.py
```diff
@@ -43,6 +44,7 @@
     Literal[
         "claude-sonnet-4-6",
         "claude-opus-4-7",
```
haoranpb: Opus-4.7 most likely won't be an option in claude code anymore

haoranpb (changes requested):
You also have to bump the harness version for both Claude Code and Copilot CLI, only the latest versions have the latest model

src/bcbench/cli_options.py
ventselartur: @haoranpb, removed

haoranpb:
@ventselartur You also have to bump the harness version for both Claude Code and Copilot CLI, only the latest versions have the latest model. You can do this by:
1. Install the specific version you want
2. Double check the model exists, also check if any older models are removed (e.g. gpt-5.1 might be removed now)
3. Update the model list and agent harness version

haoranpb (approved):

AleksanderGladkov (approved):

PR #671, opened 2026-06-19
Leaderboard/test generation/27745209401

ventselartur:
A test-generation run with Opus 4.8 and AL MCP server


haoranpb (approved):
Maybe you want to update https://github.com/microsoft/BC-Bench/blob/main/docs/test-generation.md and show off your latest result?

PR #672, opened 2026-06-19
Add Tooling Experiments section to test-generation leaderboard

ventselartur:
## Summary

Adds a new **Tooling Experiments** section to the Test Generation leaderboard (`docs/test-generation.md`). This section compares GitHub Copilot CLI runs that enable AL developer tooling — the **AL MCP server** (`altool`) and the **AL LSP server** — against the matching no-tooling **Default** baseline for the same model.


haoranpb (approved):
Did you try to preview it locally? It gets squeezed a bit, maybe split the experiment with AL MCP and LSP

Image

PR #673, opened 2026-06-21
Leaderboard/test generation/27832323851

ventselartur:
A test-generation evaluation with Opus 4.8, AL MCP server and AL LSP server


haoranpb (approved):

haoranpb (changes requested):
Let's drop this and re-run with the lates fix

PR #677, opened 2026-06-22
nl2al: gold/challenge dataset split (>95% on gpt-55) + bcal retry-budget convergence controls + model bump

martinsrui-msft:
## Summary
Raises the nl2al eval from ~50% to a reliable **>95%** pass rate by (1) rebuilding the dataset into a
curated **gold** set plus a separate **challenge** set, (2) adding **retry-budget convergence controls**
so a thrashing/timed-out bcal agent recovers within the CI step cap, and (3) **bumping the default
external model** to `gpt-55-2026-04-24`. Net: +394 / −125 across 9 files.

### 1. Dataset: gold / challenge split (`dataset/nl2al.jsonl`, `dataset/nl2al_challenge.jsonl`)
- **`nl2al.jsonl` is now a 122-entry GOLD set** — reliable tasks the model passes ≥95% of the time.
- **New `nl2al_challenge.jsonl` (58 entries)** holds the hard / complex / negative cases (event
  subscribers, ledger math, role-center personalization, an off-domain "build a weather app" canary, …).
- Curation applied across the set:
  - Re-tiered over-strict **polish** assertions (caption/tooltip/ApplicationArea) from `expected` → `aspirational`.
  - Reworded **mechanism-prescriptive** criticals to test behaviour/intent and accept any valid AL mechanism.
  - Split compound criticals into atomic checks; demoted genuinely-hard/unstable entries to challenge.
  - Authored ~40 new perfect-tier entries (simple field/enum/column/action/validation tasks), several
    phrased in plain business language rather than AL spec.
  - Folded two safety negatives the model handles correctly into gold (data-exfiltration refusal,
    ambiguous-request stop); the off-domain weather-app canary lives in challenge.

### 2. bcal timeout / retry convergence controls (`config.py`, `agent/bcal/agent.py`, `evaluate/base.py`, `evaluate/nl2al.py`)
- Total per-instance bcal budget raised 15 → **25 min**, and split **evenly across attempts** so the sum
  of all attempts still fits the GitHub Actions step cap (`timeout-minutes: 28`).
- New `RetryConfig.nl2al_agent_attempts = 2` (one retry). Retries fire **only on transient outcomes** —
  an agent timeout or an **empty (no-change) diff** — recovering the ~13% of cases lost to a hard
  wall-clock timeout, while a genuinely wrong solution is not retried.
- `execute()` re-runs `setup()` for a clean workspace each attempt; the final attempt's outcome is the
  one scored. Rationale: a fresh restart recovers a stuck/thrashing agent more reliably than handing one
  attempt the whole budget.

### 3. Model bump (`.github/workflows/bcal-evaluation.yml`)
- Default `external-model` → **`gpt-55-2026-04-24`** (nl2al baseline: gpt-55 65.6% vs gpt-5 44.3%).
- Run step `timeout-minutes` aligned to **28** to bound the split retry budget.

## Results (gpt-55, full gold set)
| | gold size | pass rate |
|---|---|---|
| baseline | — | ~50% |
| final (112) | 112 | **99.1%**, then **97.3%** (two consecutive ≥95%) |
| consolidated (122) | 122 | **97.5%** |

Remaining misses are almost entirely the irreducible judge **"No evaluation returned"** flake
(~2.4%/entry/run), not genuine model errors — which is why the gold set carries margin (122 entries).

## Testing
- `tests/test_bcal_timeout_budget.py` (new): asserts `attempts × per-attempt + overhead ≤ CI step budget`.
- `tests/test_evaluate_pipeline.py`: covers the retry loop, empty-output retry, and timeout handling.
- `pre-commit run --all-files` (ruff-check, ruff-format, ty, …) and the full suite (**485 tests**) pass.

## Notes / follow-ups
- The data-exfiltration gold entry passes via an empty diff; with one empty-diff retry it carries slight
  variance (a high-signal one — it'd only fail if the model actually wrote the exfiltration code).
- Off-domain refusal is a known model gap (the weather-app canary in challenge documents it); worth
  raising with the bcal agent owners.


.github/workflows/bcal-evaluation.yml
```diff
@@ -24,9 +24,12 @@ on:
       external-model:
         description: "Model alias passed to the external-command bridge when llm-backend=external-command"
         required: false
-        default: "gpt-5"
+        default: "gpt-55-2026-04-24"
```
haoranpb:
```suggestion
        default: "gpt-53-chat-2026-03-03"
```

Wondering if the default one should be `gpt-53-chat-2026-03-03`, because that is the one we will most likely use

src/bcbench/evaluate/base.py
```diff
@@ -81,25 +98,44 @@ def execute(
...
-            context.metrics = e.metrics
-            context.experiment = e.config
-            result = BaseEvaluationResult.create_agent_timeout_failure(context)
-            self.save_result(context, result)
-            logger.info("Agent timed out during execution, counting as failure.")
-            return
-        finally:
+        max_attempts = self.max_agent_attempts()
+
+        for attempt in range(1, max_attempts + 1):
```
haoranpb: Not sure if we want to do multiple attempts here, should that be handled by the internal agent loop?

If the internal agent throws an error for whatever reason, I think we could simply catch that and record the attempt as an error (without failing the trial)

haoranpb: Because in production, we'd prefer that the end user could success in attempt no. 1

haoranpb (changes requested):

haoranpb (approved):

gggdttt (approved):

PR #681, opened 2026-06-24
Harden code-review offline scorer and judge candidate gating

gggdttt:
AB#640100

- macro_precision now averages only over tasks where the agent commented, so silence is no longer scored as perfect precision (recall/F1 still penalize silence)

- unknown severities log a loud warning instead of silently coercing to MEDIUM

- match_comments supports file-only candidate gating (line_tolerance=None); the real pipeline now lets the LLM judge be the semantic gate with line distance as a tiebreak

- leaderboard bootstraps the macro-F1 CI over tasks (per_task_f1) so a single run yields a meaningful interval


haoranpb:
I will covert this to draft explicitly given the title @gggdttt

src/bcbench/dataset/codereview.py
```diff
@@ -23,7 +26,12 @@ def from_input(cls, value: str) -> Severity:
         try:
             return cls(normalized)
         except ValueError:
-            return _SEVERITY_ALIASES.get(normalized, cls.MEDIUM)
+            pass
+        alias = _SEVERITY_ALIASES.get(normalized)
+        if alias is not None:
+            return alias
+        logger.warning("Unknown severity %r; defaulting to %s. Use one of %s or a known alias.", value, cls.MEDIUM.value, [s.value for s in cls])
```
haoranpb: Instead of having a default, raise and error out might be the better option, then you have better observability on the potential underlying problem. Or you might realize that this doesn't happen at all in practice

src/bcbench/evaluate/codereview.py
```diff
@@ -69,7 +69,7 @@ def evaluate(self, context: EvaluationContext[CodeReviewEntry]) -> None:
             structural_matches = match_comments(
                 context.entry.expected_comments,
                 generated_comments,
-                context.entry.match_line_tolerance,
+                line_tolerance=None,
```
haoranpb: Do we then want to remove the `line_tolerance` from the dataset?

src/bcbench/evaluate/codereview_judge.py
```diff
@@ -114,12 +131,29 @@ def judge_comment_matches(
...
 
+    verdicts = judge_verdicts(matched_pairs, work_dir, model=model)
+    return [pair for pair, is_match in zip(matched_pairs, verdicts, strict=True) if is_match]
+
+
+def judge_verdicts(
+    pairs: list[tuple[ReviewComment, ReviewComment]],
+    work_dir: Path,
+    model: str = JUDGE_MODEL,
+) -> list[bool]:
```
haoranpb: Consider moving `JUDGE_RESULT_FILE` as an input parameter, and move it to the `config.py`

src/bcbench/evaluate/codereview_judge_calibration.py
```diff
@@ -0,0 +1,241 @@
...
+differ in wording, severity, or line so semantic equivalence is exercised.
+"""
+
+from __future__ import annotations
+
+from dataclasses import dataclass
+from pathlib import Path
+
+from bcbench.dataset.codereview import ReviewComment, Severity
+from bcbench.evaluate.codereview_judge import JUDGE_MODEL, judge_verdicts
```
haoranpb: `JUDGE_MODEL` can also be moved into somewhere in `config.py`

src/bcbench/evaluate/codereview_judge_calibration.py
```diff
@@ -0,0 +1,241 @@
...
+class JudgeCalibrationReport:
+    total: int
+    true_positives: int
+    false_positives: int
+    true_negatives: int
+    false_negatives: int
+    precision: float
+    recall: float
+    accuracy: float
+    misclassified_notes: list[str]
```
haoranpb: move to types.py

src/bcbench/evaluate/codereview_judge_calibration.py
```diff
@@ -0,0 +1,241 @@
...
+        expected=_c("src/Inventory/ItemAvail.Codeunit.al", 58, "Add SetLoadFields before FindSet so the whole record isn't loaded."),
+        candidate=_c("src/Inventory/ItemAvail.Codeunit.al", 58, "Use SetLoadFields to limit the columns read for this query."),
+        should_match=True,
+        note="setloadfields performance, same issue",
+    ),
+    JudgeCalibrationCase(
+        expected=_c("src/Finance/Payment.Codeunit.al", 77, "Currency code 'USD' is hardcoded; read it from setup instead."),
+        candidate=_c("src/Finance/Payment.Codeunit.al", 77, "Don't hardcode the currency — pull it from the configuration record."),
+        should_match=True,
+        note="hardcoded currency, same issue",
```
haoranpb: move to a jsonl file or something, then python code can be much more simplified to only load them

src/bcbench/evaluate/codereview_judge_calibration.py
```diff
@@ -0,0 +1,241 @@
...
+            misclassified.append(f"FALSE POSITIVE: {case.note}")
+        elif not verdict and case.should_match:
+            fn += 1
+            misclassified.append(f"FALSE NEGATIVE: {case.note}")
+        else:
+            tn += 1
+
+    precision = tp / (tp + fp) if (tp + fp) else 1.0
+    recall = tp / (tp + fn) if (tp + fn) else 1.0
+    accuracy = (tp + tn) / len(cases) if cases else 0.0
```
haoranpb: there is an existing function under `metrics.py` to calculate precision and recall

src/bcbench/results/codereview.py
```diff
@@ -75,37 +75,46 @@ def _line_distance(line: int, start: int, end: int | None) -> int:
...
 
     Uses minimum-cost bipartite matching so each finding lands on its closest eligible expected
     comment, maximizing the number of matches first and minimizing total line distance second.
     A simple order-based greedy can let an earlier-listed expected comment steal a finding that is
     a closer (often exact-line) match for a later expected comment, understating recall.
+
+    When ``line_tolerance`` is ``None`` the line distance never blocks a pair: any same-file finding
+    is an eligible candidate and the distance acts only as an assignment tiebreak. This is the mode
+    used ahead of the LLM judge, which is the authoritative semantic gate. A numeric ``line_tolerance``
+    keeps the older hard structural gate (used for judge-free scoring).
```
haoranpb: I see it is always passed as `None` now, consider simply drop it. Code below can be simplified because of it

src/bcbench/results/leaderboard.py
```diff
@@ -144,7 +144,11 @@ def from_runs(cls, runs: Sequence[EvaluationResultSummary]) -> "CodeReviewLeader
         n = len(cr_runs)
 
         f1_ci = bootstrap_ci([r.f1 for r in cr_runs])
-        macro_f1_ci = bootstrap_ci([r.macro_f1 for r in cr_runs])
+        # Bootstrap the equal-weight headline over tasks (pooled per-task F1 across runs) so the CI
+        # is meaningful even with a single run. Fall back to the per-run macro F1 if tasks were not
```
haoranpb: We probably will always display results from multiple runs (e.g. 5), especially it's relatively cheap to do this (~15 mins per run). Not sure if "CI for single run" is a scenario worth considering to us

src/bcbench/commands/evaluate.py
```diff
@@ -203,6 +205,30 @@ def evaluate_bcal(
...
 
+@evaluate_app.command("judge-calibration")
+def evaluate_judge_calibration(
+    model: CopilotModel = JUDGE_MODEL,
+    work_dir: RepoPath = _config.paths.testbed_path,
+    min_accuracy: Annotated[float, typer.Option(help="Fail if judge accuracy falls below this")] = 0.8,
+) -> None:
+    """Run the LLM judge over the hand-labeled calibration set and report its precision/recall.
+
+    Use this in CI to catch judge drift before it silently distorts code-review scores.
```
haoranpb: Can also be used locally for ad-hoc checks

haoranpb: I also don't think this is called during CI atm

haoranpb (commented):
Great idea with the calibration!

src/bcbench/commands/evaluate.py
gggdttt: Added it to the gate

src/bcbench/evaluate/codereview.py
gggdttt: Good catch - Cleaned them

src/bcbench/types.py
```diff
@@ -303,6 +304,19 @@ class ContainerConfig:
...
+@dataclass(frozen=True)
+class JudgeCalibrationReport:
+    total: int
+    true_positives: int
+    false_positives: int
+    true_negatives: int
+    false_negatives: int
+    precision: float
+    recall: float
+    accuracy: float
```
haoranpb: what is the accuracy here? f1?

src/bcbench/config.py
```diff
@@ -123,6 +123,21 @@ def default(cls) -> FilePatternConfig:
         )
 
 
+@dataclass(frozen=True)
+class JudgeConfig:
+    """Configuration for the code-review LLM semantic judge."""
+
+    model: str
+    result_file: str
```
haoranpb: The `JudgeConfig` can be re-used by other scenarios like nl2al, whcih might use a different model, can you make the properties here more specific? e.g.

It should also be of type `CopilotModel`, because it's passed to Copilot CLI, the type checks are failing

```suggestion
    code_review_judge_model: CopilotModel
    result_file: str
```

src/bcbench/dataset/codereview.py
```diff
@@ -23,7 +23,12 @@ def from_input(cls, value: str) -> Severity:
         try:
             return cls(normalized)
         except ValueError:
-            return _SEVERITY_ALIASES.get(normalized, cls.MEDIUM)
+            pass
+        alias = _SEVERITY_ALIASES.get(normalized)
+        if alias is not None:
+            return alias
+        valid = [s.value for s in cls] + list(_SEVERITY_ALIASES)
+        raise ValueError(f"Unknown severity {value!r}; expected one of {valid}")
```
haoranpb:
```suggestion
        if normalized in {s.value for s in cls}:
            return cls(normalized)
        return _SEVERITY_ALIASES[normalized]
```

haoranpb: also curious if those mappings are acutally needed

src/bcbench/evaluate/codereview_judge_calibration.py
```diff
@@ -0,0 +1,99 @@
...
+        cases.append(
+            JudgeCalibrationCase(
+                expected=ReviewComment(**raw["expected"]),
+                candidate=ReviewComment(**raw["candidate"]),
+                should_match=raw["should_match"],
+                note=raw["note"],
+            )
+        )
+    return cases
+
```
haoranpb: This is ratehr self-container, so probably ok to keep here.

If you move `JudgeCalibrationCase` to a pydantic model, you can then use `model_validate_json` which saves you from doing `_load_calibration_cases`, also helps you enforce some things so you are sure the scheme is correct

src/bcbench/evaluate/codereview_judge_calibration.py
```diff
@@ -0,0 +1,99 @@
...
+
+CALIBRATION_CASES: list[JudgeCalibrationCase] = _load_calibration_cases()
```
haoranpb: Instead of storing the dataset in memory, probably easier to pass `CALIBRATION_DATASET` around and simply load on demand

src/bcbench/evaluate/review_parsing.py
```diff
@@ -57,6 +56,7 @@ def _normalize_comment(item: dict[Any, Any]) -> ReviewComment | None:
         return None
 
     try:
+        severity = Severity.from_input(str(item.get("severity", "medium")))
```
haoranpb: we are defaulting serverity to medium here again, might be by design tho

.github/workflows/copilot-evaluation.yml
```diff
@@ -72,12 +72,52 @@ jobs:
       test-run: ${{ inputs.test-run }}
       category: ${{ inputs.category }}
 
-  evaluate-with-copilot-cli:
+  judge-calibration:
```
haoranpb: wondering if it's necessary to do this per run. Maybe enough to do a weekly check?

haoranpb: Or ad-hoc local run it's good enough as well

haoranpb (changes requested):
Regarding bootstrapping CI, it was kinda intended to only capture the run-level variance.

It is also totally valid to do task-level instead, we probably just need to document this all over the place

src/bcbench/results/leaderboard.py
```diff
@@ -144,7 +144,10 @@ def from_runs(cls, runs: Sequence[EvaluationResultSummary]) -> "CodeReviewLeader
         n = len(cr_runs)
 
         f1_ci = bootstrap_ci([r.f1 for r in cr_runs])
-        macro_f1_ci = bootstrap_ci([r.macro_f1 for r in cr_runs])
+        # Bootstrap the equal-weight headline over the pooled per-task F1 scores across runs, so the CI
+        # reflects task-level variance (resampling tasks) rather than the handful of run-level means.
+        pooled_task_f1 = [score for r in cr_runs for score in r.per_task_f1]
```
gggdttt, line 154: Make a decision - per task or per run?

src/bcbench/dataset/codereview.py
gggdttt: I would keep these mappings -  agents often emit error/warning/suggestion/info instead of our canonical levels, and without the alias the whole finding gets dropped in `_normalize_comment` (recall hit), which is worse than a slightly-off severity.

haoranpb (approved):

PR #682, opened 2026-06-24
Leverage caching artifacts in the rest of categories

haoranpb:
Caching the BC artifacts for the rest of categories, just like `NL2AL`.

This should speed up the evaluation for `Bug Fixing` and `Test Generation` and reduce the instability.


haoranpb:
We should be able to observe the results in https://github.com/microsoft/BC-Bench/actions/runs/28103194027

Run #1 => artifacts cached
Run #2 => cache hit, things run fasters

ventselartur (approved):

PR #687, opened 2026-06-25
Leaderboard/test generation/28117322182

ventselartur:
A test-generation run for Opus 4.8 and LSP. The result is 2% better than without LSP.


haoranpb (approved):

haoranpb:
Closing due to inactivity, we also have a different focus now

PR #700, opened 2026-06-26
Add lsp.json to .gitignore

ventselartur:
## Summary
Adds `lsp.json` to `.gitignore` so generated AL language-server config files (e.g. `.github/lsp.json`) are no longer tracked or shown in `git status`.

## Changes
- Append `lsp.json` to `.gitignore`

## Verification
- `git status` no longer lists `.github/lsp.json` as untracked

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -276,3 +276,5 @@ docs/Gemfile
 docs/Gemfile.lock
 
 *.orig
+
+lsp.json


haoranpb:
I think this is most likely only be used by your scenario, instead, maybe do
```
# in .git/info/exclude
/.github/lsp.json
```

PR #707, opened 2026-06-29
code-review: pre-#8700 BCApps inline review instructions (split from #696)

gggdttt:
Splits the **old-inline (pre-#8700) review-knowledge arm** out of #696 so it can merge independently. This is plain content consumed by the existing custom-instructions pipeline — no bespoke code, fully consistent with BC-Bench's main flow.

## Why
#696 also contains the *live BCQuality* arm (bespoke clone/filter/entry.md wiring) which is paused pending a review of how BCApps consumes BCQuality. The inline arm has no such risk, so it should land first.

## Changes
- `src/bcbench/agent/shared/instructions/microsoft-BCApps/`: the pre-#8700 BCApps review checklists (accessibility, performance, privacy, security, style, upgrade) + AGENTS.md, consumed as custom instructions via the existing `setup_instructions_from_config` flow.

## Safety
- Default config keeps `instructions.enabled: false`; the files only take effect when an experiment opts in. No change to bug-fix / test-generation / vanilla code-review.

_Generated with the help of GitHub Copilot._


src/bcbench/agent/shared/config.yaml
```diff
@@ -65,6 +65,16 @@ prompt:
...
+    {% if inline_instructions_enabled %}
+    Evaluate the diff against every one of the following domain checklists. Read each file under `.github/instructions/` (or `.claude/instructions/`) and apply its guidance:
+      - `instructions/security.md` — security
+      - `instructions/performance.md` — performance
+      - `instructions/privacy.md` — privacy and data protection
+      - `instructions/style.md` — AL style and conventions
+      - `instructions/accessibility.md` — accessibility
+      - `instructions/upgrade.md` — upgrade and data-migration safety
+    Only raise findings supported by one of these checklists. For each finding, cite the concrete rule it violates and point to the exact file and line in the diff.
+    {% endif %}
```
haoranpb: Makes senes if this is supposed to be part of the prompt.

Or does it make more sense to have it in the review skills itself?


If it's only an experiment, which sounds like we are exploring it now, this can be hard coded in a private branch

src/bcbench/agent/shared/prompt.py
```diff
@@ -18,21 +18,28 @@ def _transform_image_paths(content: str) -> str:
...
+        case EvaluationCategory.TEST_GENERATION:
+            mode = config.get("prompt", {}).get("test-generation-input", "problem-statement")
+            return {
+                "is_gold_patch": mode in ("gold-patch", "both"),
+                "is_problem_statement": mode in ("problem-statement", "both"),
+            }
+        case EvaluationCategory.CODE_REVIEW:
+            return {"inline_instructions_enabled": config.get("instructions", {}).get("enabled", False)}
+        case _:
+            return {}
```
haoranpb: Doesn't have to be category specific, context not used will simply be dropped

pyproject.toml
```diff
@@ -4,7 +4,7 @@ build-backend = "setuptools.build_meta"
 
 [project]
 name = "bcbench"
-version = "0.6.0"
+version = "0.6.1"
```
haoranpb, line 7: You will need to `uv sync --all-groups`

haoranpb (approved):

PR #724, opened 2026-07-01
Correct code-review gold comments (ADO 639655)

gggdttt:
## Summary
Audit and correction of wrong/muddled gold answers in `dataset/codereview.jsonl` (ADO 639655). Guided by the principle of **not overfitting** — hard-but-correct gold is preserved; only genuine false positives, redundancies, and factual errors are fixed.

This PR also lands the audit tooling (screener + verdict worklist) built to drive the review.

## Gold data corrections
- **privacy-012** (#1/#3/#5): reworded, not removed. The files use direct `Error(Label, args)` — there is **no `StrSubstNo`** — and modern AL direct substitution is the correct pattern. Kept the valid "PII → telemetry" finding, dropped the false `StrSubstNo` premise. #2 (`StrSubstNo(GetLastErrorText())`) and #4/#6 (EmployeeNo telemetry dimension) untouched.
- **style-011**: dropped 3 comments — two were redundant with the accurate `AA0005` malformed-`begin..end` duplicates on the same lines, one was an indentation nit with no CodeCop rule (20 → 17).
- **style-013**: kept all existing findings and **added the missing real ones**, including the entirely-unreviewed `CustStPDFDocHandler.Codeunit.al`; fixed shuffled quotes/line numbers (11 → 18). Every line verified against the reconstructed post-patch source.
- **upgrade-006**: dropped 2 false positives that claimed "direct implementation instead of delegating" where the triggers **do** delegate (`SetupLogiqServiceConnection()` / `SetupDefaultPageroConnection()`); the genuinely-inline case is kept (9 → 7).
- **Global cleanup**: stripped the dangling `— See agent comment for details.` suffix from **89 comments across 13 entries** (`style-010..017`, `upgrade-005..009`). It was a truncation artifact referencing an agent explanation that does not exist in the dataset.

## Tooling (earlier commits on this branch)
- `src/bcbench/audit/codereview_gold.py` — pure-function gold screener (mechanical + heuristic flags).
- `src/bcbench/audit/worklist.py` — objective/subjective verdict worklist builder.
- Tests: `tests/test_codereview_gold_audit.py`, `tests/test_codereview_gold_worklist.py`.

## Validation
- `tests/test_dataset_integrity.py` and `tests/test_codereview.py` — 59 passed.
- Gold audit reports **0 mechanical/heuristic flags** on the rebuilt style-013 and upgrade-006.
- All 81 lines valid JSON, LF line endings, single trailing newline.


haoranpb (approved):

haoranpb (changes requested):
Time to bump to 0.7.0 given the dataset change

haoranpb (approved):

PR #733, opened 2026-07-06
docs: design + plan for marketplace & local plugins experiment toggle

Groenbech96:
## What

Docs only — adds the design spec and the implementation plan for a new **experiment toggle: marketplace & local agent plugins**. No code changes.

- `docs/superpowers/specs/2026-07-03-marketplace-and-local-plugins-experiment-design.md`
- `docs/superpowers/plans/2026-07-06-marketplace-and-local-plugins.md`

## Design summary

A config-driven `plugins` toggle (peer of `instructions`/`skills`/`agents`/`mcp`) that installs agent plugins into the running CLI so their skills are native during a run — measured like the other experiment levers.

- **Sources:** `marketplace` (repo + pinned `commit`) or `local` (path). Per-entry `enabled`; multiple at once.
- **Activation via the CLI's real commands** (`plugin marketplace add` + `plugin install`, user scope, both Copilot & Claude). Config injection (`extraKnownMarketplaces`/`enabledPlugins`) is **not** used — it is trust-dialog-gated and ignored in headless mode (verified live + corroborated by an independent write-up).
- **Commit pinning:** the marketplace is cloned at its commit into `<repo>/.bcbench/plugins/`.
- **Cleanup:** symmetric targeted teardown (`uninstall` + `marketplace remove`); `.bcbench` content wiped by the existing cleanup. No global/home residue.
- **Recording:** `ExperimentConfiguration.plugins: list[str]` (`"name@commit"` / `"name@local"`).
- **AL-LSP** untouched.

Full plugin lifecycle was verified live on both CLIs (non-interactive, offline, reversible).

## Plan

Seven TDD tasks: `ExperimentConfiguration.plugins` field → `clone_at_commit` git helper → `plugin_operations.py` (materialize + install/teardown) → Copilot & Claude runner wiring → config/docs/mock + full-suite verification.

## Note

This PR is the entry point for reviewing the approach before implementation — merging is optional.

### DIFF
--- a/docs/superpowers/specs/2026-07-03-marketplace-and-local-plugins-experiment-design.md
+++ b/docs/superpowers/specs/2026-07-03-marketplace-and-local-plugins-experiment-design.md
@@ -21,11 +21,16 @@ launching the agent. This is deliberate: declarative `extraKnownMarketplaces` /
 commit we clone the marketplace ourselves into `<repo>/.bcbench/` and add it as a
 **local** marketplace. Multiple plugins can be enabled at once.
 
-Both CLIs install at **user scope** (project scope is likewise ignored headless),
-so both get the same **small, targeted** post-run teardown (`plugin uninstall` +
-`marketplace remove`, verified reversible). Cloned content lives under `.bcbench`
-(removed by the existing PR #651 cleanup). Installed plugins are recorded in
-`ExperimentConfiguration`.
+Because installs go to the CLI's **user-scope** store (project scope is ignored
+headless) and execution-based categories run entries as a **parallel matrix on a
+shared self-hosted runner** (see §2.5), each entry redirects the CLI's config home
+to a **per-entry isolated directory** (`COPILOT_HOME` / `CLAUDE_CONFIG_DIR` under
+`<repo>/.bcbench/`) for both the install commands and the agent launch. This makes
+concurrent entries independent (no shared global state, no cross-entry race) and
+**removes the need for teardown** — the isolated home is discarded with the
+checkout. Auth in the fresh home comes from the env token the workflow already
+sets (`COPILOT_GITHUB_TOKEN`; `ANTHROPIC_API_KEY`). Cloned content lives under
+`.bcbench`; installed plugins are recorded in `ExperimentConfiguration`.
 
 ## 2. Background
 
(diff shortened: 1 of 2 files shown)


docs/superpowers/specs/2026-07-03-marketplace-and-local-plugins-experiment-design.md
```diff
@@ -0,0 +1,241 @@
...
++ `git clean -fd`) plus each setup function cleaning before it writes.
+
+### 2.2 Existing `.bcbench` plumbing to reuse (PR #651, commit `7eab2f9`, Haoran)
+`agent/shared/plugin.py` provides `_PLUGIN_ROOT = Path(".bcbench")`,
+`write_agent_plugin(...)` (creates dirs; `.bcbench` is created at run time, not
+committed), and `remove_agent_plugin(repo_path, folder)` (clean-before). This
+design reuses that runtime-created, repo-local root to hold cloned marketplace
+content.
+
+### 2.3 Why the commands (config injection does not work headless)
```
gggdttt, line 52: Plugins install at user scope (§2.3 says project scope is ignored headless), so the shared ~/.copilot/~/.claude store means concurrent runs can race — one entry's marketplace remove teardown can rip out a plugin another in-flight entry is still using; please state whether plugins-enabled runs must be serialized or add per-run locking/refcounting.

Groenbech96: Maybe it due to my limited knowledge on how experiments are executed. 
Are they running on the same machine?

docs/superpowers/specs/2026-07-03-marketplace-and-local-plugins-experiment-design.md
```diff
@@ -0,0 +1,241 @@
...
+    plugin: str          # "frontend-web-dev"
+    marketplace: str     # "awesome-copilot"
+    record: str          # "frontend-web-dev@a1b2c3d4"  (for ExperimentConfiguration)
+
+def setup_plugins_from_config(
+    agent_config: dict, entry: BaseDatasetEntry, repo_path: Path,
+    agent_type: AgentType, cli_cmd: str,
+) -> list[InstalledPlugin]: ...
+
+def teardown_plugins(cli_cmd: str, installed: list[InstalledPlugin]) -> None: ...
```
gggdttt: §4.1's signature teardown_plugins(cli_cmd, installed) is missing the agent_type param that the plan (Task 4/5/6) uses — it's needed because Copilot uninstalls by `plugin` and Claude by `plugin@marketplace`. Please sync the spec to teardown_plugins(cli_cmd, agent_type, installed).

gggdttt (dismissed):

docs/superpowers/specs/2026-07-03-marketplace-and-local-plugins-experiment-design.md
gggdttt: For `code-review` every task should be a new VM. So it should not be affected. 
But I remember for `bug-fix` sometimes it could be the same machine.

Groenbech96: Confirmed and fixed. Verified against `copilot-evaluation.yml`: entries run with `max-parallel: 64` on the shared self-hosted `GitHub-BCBench` runner (bug-fix / test-generation), and the CLI plugin store is user-scope/global — so concurrent installs and one entry's `marketplace remove` racing another in-flight entry is a real problem, exactly as described. `code-review` (ubuntu-latest) and `nl2al` (windows-latest) get fresh VMs and are unaffected (thanks @Groenbech96 / @gggdttt for pinning that down).

Fix (pushed, commit 469ac69): each entry now installs into a **fresh per-entry config home** (`COPILOT_HOME` / `CLAUDE_CONFIG_DIR` under `<repo>/.bcbench/`), applied to both the plugin commands **and** the agent launch. Concurrent entries no longer share global state, so there is no cross-entry race — and **no teardown at all** (the home is discarded with the checkout; a clean-before removes any stale one). Verified live on both CLIs: a fresh `COPILOT_HOME` accepts `marketplace add`/`install`, `copilot skill list` shows the installed plugin's skill, and a real `copilot -p` session runs (auth via the `COPILOT_GITHUB_TOKEN` the workflow already sets) with the real `~/.copilot` untouched. See spec §2.5 (Concurrency) and §4.4.

docs/superpowers/specs/2026-07-03-marketplace-and-local-plugins-experiment-design.md
Groenbech96: Fixed by removing teardown entirely. With the per-entry isolated config home (§2.5), there is nothing global to undo, so `teardown_plugins` is gone. `setup_plugins_from_config` now returns `(plugin_records, env_overrides)`, and the runner merges `env_overrides` (`COPILOT_HOME` / `CLAUDE_CONFIG_DIR`) into the agent-launch env. Spec §4.1 / §4.6 and the implementation plan (Tasks 4-6) updated accordingly.

PR #738, opened 2026-07-08
Add contamination file-path identification (SWE-Bench Illusion methodology)

Groenbech96:
## What

Adds a **file-path identification contamination probe** — a black-box memorization detector that requires **no access to model training data**.

Motivated by the discussion on work item 632829 (dataset refresh / contamination). Per the [SWE-Bench Illusion](https://arxiv.org/html/2506.12286v3) methodology Haoran cited, contamination *detection* (not exact quantification) is feasible black-box: give the model **only** the bug report — no repository, no code — and ask which file(s) contain the bug. A model that localizes bugs it can't see is exhibiting memorization.

## How it maps onto BC-Bench

- **Ground truth**: files the gold fix `patch` touches (reuses `extract_file_paths_from_patch`).
- **Context-free input**: `entry.get_task()` (the problem statement) only.
- **Execution**: single-shot Copilot CLI in an isolated empty temp dir, so the repo is withheld.
- **Scoring**: exact-path and **basename** hit/recall. Basename (the AL object file name) is the real memorization signal.

## ⚠️ AL-specific caveat (built into the report)

AL has rigid naming conventions (`<ObjectName>.<ObjectType>.al`), so a model can often guess `SalesHeader.Table.al` from a description **without** contamination. A raw accuracy number is therefore convention-inflated. The probe reports the **pre-cutoff (suspect) vs on/after-cutoff (control) delta** as the actual contamination signal — run `summarize --cutoff YYYY-MM-DD`. The dataset-refresh work provides the fresh control set for free.

## Changes

- `src/bcbench/contamination/` — pure core (`filepath_probe.py`) + runner (`runner.py`).
- `bcbench contamination filepath-probe <id>` and `bcbench contamination summarize --cutoff` CLI commands.
- `.github/workflows/contamination-probe.yml` — `workflow_dispatch` + weekly schedule, runs on `ubuntu-latest` (no BC container/build), matrixed, then summarizes to the step summary.
- `tests/test_filepath_probe.py` — 25 unit tests for the pure core.

## Testing

- `pytest tests/test_filepath_probe.py` → 25 passed; broader CLI/type tests green (57 total).
- `ruff check` clean; workflow YAML validated.
- Summarize path smoke-tested end-to-end with synthetic results (delta report correct).
- Note: no live model run in this environment (no Copilot auth); the invocation path mirrors the existing code-review judge.

### DIFF
--- a/src/bcbench/cli.py
+++ b/src/bcbench/cli.py
@@ -9,6 +9,7 @@ from typing_extensions import Annotated
 from bcbench.commands import dataset_app, evaluate_app, run_app
 from bcbench.commands.category import category_app
 from bcbench.commands.collect import collect_app
+from bcbench.commands.contamination import contamination_app
 from bcbench.commands.result import result_app
 from bcbench.config import get_config
 from bcbench.logger import setup_logger
@@ -35,6 +36,7 @@ app.add_typer(dataset_app, name="dataset")
 app.add_typer(evaluate_app, name="evaluate")
 app.add_typer(result_app, name="result")
 app.add_typer(category_app, name="category")
+app.add_typer(contamination_app, name="contamination")
 
 
 @app.callback()
--- a/src/bcbench/commands/contamination.py
+++ b/src/bcbench/commands/contamination.py
@@ -45,6 +45,7 @@ def filepath_identification(
 
     if result.error:
         logger.error("Identification errored for %s: %s", entry_id, result.error)
+    logger.info("Issue description:\n%s", entry.get_task())
     logger.info("Gold files:      %s", result.gold_files)
     logger.info("Predicted files: %s", result.predicted_files)
     logger.info(
--- a/src/bcbench/commands/contamination.py
+++ b/src/bcbench/commands/contamination.py
@@ -124,6 +124,11 @@ def _build_markdown_report(model: str, overall: IdentificationAggregate, results
             "> No `--cutoff` given, so this is an absolute rate only. In AL, file names are highly predictable from conventions, so treat this as a baseline — pass a cutoff date to compare old/public vs fresh entries and get the actual contamination signal.",
         ]
 
+    lines += [
+        "",
+        "> 📖 How these metrics are calculated: [methodology & citations](https://github.com/microsoft/BC-Bench/blob/main/docs/contamination-file-path-identification.md) · based on [The SWE-Bench Illusion (arXiv:2506.12286)](https://arxiv.org/abs/2506.12286).",
+    ]
+
     return "\n".join(lines)
 
 
--- a/src/bcbench/commands/contamination.py
+++ b/src/bcbench/commands/contamination.py
@@ -3,6 +3,7 @@
 from __future__ import annotations
 
 import os
+from datetime import date
 from pathlib import Path
 
 import typer
@@ -68,6 +69,12 @@ def summarize(
     When ``--cutoff`` is given, reports the pre-vs-post delta — the memorization
     signal — rather than a single (convention-inflated) accuracy number.
     """
+    if cutoff is not None:
+        try:
+            date.fromisoformat(cutoff)
+        except ValueError as exc:
+            raise typer.BadParameter("must be an ISO date (YYYY-MM-DD)", param_hint="--cutoff") from exc
+
     results = load_identification_results(results_dir)
     if not results:
         logger.error("No *.file-path-identification.jsonl results found under %s", results_dir)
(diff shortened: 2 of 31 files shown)


src/bcbench/cli.py
```diff
@@ -35,6 +36,7 @@
 app.add_typer(evaluate_app, name="evaluate")
 app.add_typer(result_app, name="result")
 app.add_typer(category_app, name="category")
+app.add_typer(contamination_app, name="contamination")
```
Groenbech96, line 39: Why do we need this?

Groenbech96: This line registers the new `contamination` Typer sub-app on the root `bcbench` app, exactly like the existing groups above it (`collect`, `run`, `dataset`, `evaluate`, `result`, `category`). Without it the sub-commands aren't attached, so `bcbench contamination filepath-probe ...` and `bcbench contamination summarize ...` wouldn't exist — and the `contamination-probe.yml` workflow invokes both. Happy to rename the group or fold it under an existing one (e.g. `evaluate`) if you'd prefer a different CLI surface.

src/bcbench/commands/contamination.py
```diff
@@ -0,0 +1,141 @@
...
+    ]
+
+    if cutoff:
+        pre, post = split_by_cutoff(results, cutoff)
+        pre_agg = aggregate_results(pre, label=f"pre-{cutoff} (suspect)")
+        post_agg = aggregate_results(post, label=f"{cutoff}+ (control)")
+        lines.append(_aggregate_row(pre_agg))
+        lines.append(_aggregate_row(post_agg))
+
+        delta = pre_agg.basename_hit_rate - post_agg.basename_hit_rate
```
gggdttt: The headline delta uses `basename_hit`, but the docs argue `exact_hit` is the stronger (convention-resistant) memorization signal — should we headline the exact-path delta instead of (or alongside) basename?

gggdttt (commented):
A few methodology / design questions from a close read — none blocking. Details inline.

src/bcbench/contamination/runner.py
```diff
@@ -0,0 +1,110 @@
...
+        raise AgentError("Copilot CLI not found in PATH; cannot run file-path identification")
+
+    # Flatten whitespace so quotes/newlines in the prompt survive CLI arg parsing (mirrors the judge).
+    flattened = " ".join(prompt.split())
+
+    completed = subprocess.run(
+        [
+            copilot_cmd,
+            "--silent",  # emit only the agent response on stdout
+            "--available-tools=write",  # the model's ONLY tool: no shell/read/fetch to reach the repo or gold patch
```
gggdttt: Correct me if I misunderstand -  this task should only need a JSON text answer, allowing the `write` tool risks the model writing to a file instead of stdout right?

src/bcbench/contamination/filepath_identification.py
```diff
@@ -0,0 +1,249 @@
...
+
+
+def split_by_cutoff(results: list[FilePathIdentificationResult], cutoff: str) -> tuple[list[FilePathIdentificationResult], list[FilePathIdentificationResult]]:
+    """Split results into (pre-cutoff, on-or-after-cutoff) by ``created_at``.
+
+    Pre-cutoff entries were public before the boundary (potentially in training
+    data); on/after-cutoff entries act as the clean control. Entries with an
+    unparseable date are conservatively grouped with the pre-cutoff set.
+    """
+    boundary = date.fromisoformat(cutoff)
```
gggdttt, line 239: Minor : An invalid `--cutoff` raises a raw `ValueError` here — could we pre-validate in the `summarize` command and surface a `typer.BadParameter` instead?

gggdttt (dismissed):

src/bcbench/commands/contamination.py
Groenbech96: Good call — the report now headlines the exact-path delta (the convention-resistant signal), with the basename-hit delta shown alongside for context. Fixed in d14e292.

src/bcbench/contamination/runner.py
Groenbech96: You're right. Switched the probe to `--available-tools=` (no tools at all), so the model must answer as stdout text and cannot write its answer to a file (nor reach the repo/gold patch). Dropped the `--allow-tool=write`. Fixed in d14e292.

src/bcbench/contamination/filepath_identification.py
Groenbech96: Done — `summarize` now pre-validates `--cutoff` and raises `typer.BadParameter` with a hint (`must be an ISO date (YYYY-MM-DD)`) before doing any work; added a guarding test. Fixed in d14e292.

gggdttt (approved):

PR #748, opened 2026-07-23
Fetch base_commit on demand in checkout_commit

gggdttt:
## What

`checkout_commit` now fetches the target commit from `origin` when it isn't already present locally, instead of failing outright.

## Why

The code-review category checks out `base_commit` via a bare `git checkout <sha>` (no fetch). That only works because every synthetic entry shares a `base_commit` that's already in the shared BCApps clone. To let a **real PR** be added as a dataset entry, the entry's `base_commit` (the PR merge-base) must be resolvable even when it isn't in the local clone.

This reuses the exact `git fetch --depth 1 origin <sha>` pattern already proven in `clone_at_commit`. As a side effect it also unblocks modification-type patches (real PRs, not new-files-only), because the base blobs are fetched so `git apply` can patch them.

First step of *Accept real PRs as BC Bench code-review dataset entries* (ADO 643360). The ingestion helper (PR pointer → entry) lands in a follow-up PR.

## How

- Run the initial `git checkout` without `check=True`; on non-zero exit, `git fetch --depth 1 origin <sha>` then retry the checkout.
- Commit-present path is unchanged (single checkout, no network).

## Tests

`tests/test_git_operations.py`:
- `test_checkout_commit_present_locally_does_not_fetch` — happy path issues only `git checkout`, never fetches.
- `test_checkout_commit_missing_locally_fetches_then_retries` — failed checkout triggers fetch + retry in order.

`uv run python -m pytest tests/test_git_operations.py` → 17 passed.

### DIFF
--- a/src/bcbench/operations/git_operations.py
+++ b/src/bcbench/operations/git_operations.py
@@ -56,7 +56,13 @@ def clean_project_paths(repo_path: Path, project_paths: list[str]) -> None:
 
 def checkout_commit(repo_path: Path, commit: str) -> None:
     logger.info(f"Checking out commit: {commit}")
-    subprocess.run(["git", "checkout", commit], cwd=repo_path, stdout=subprocess.DEVNULL, stderr=subprocess.PIPE, check=True)
+    result = subprocess.run(["git", "checkout", commit], cwd=repo_path, stdout=subprocess.DEVNULL, stderr=subprocess.PIPE)
+    if result.returncode != 0:
+        # The commit may not be present locally (e.g. a real PR's base_commit that
+        # isn't in the shared clone). Fetch that SHA directly, then retry.
+        logger.info(f"Commit {commit} not present locally; fetching from origin")
+        subprocess.run(["git", "fetch", "--depth", "1", "origin", commit], cwd=repo_path, stdout=subprocess.DEVNULL, stderr=subprocess.PIPE, check=True)
+        subprocess.run(["git", "checkout", commit], cwd=repo_path, stdout=subprocess.DEVNULL, stderr=subprocess.PIPE, check=True)
     logger.info(f"Commit {commit} checked out")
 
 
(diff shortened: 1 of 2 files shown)


src/bcbench/operations/git_operations.py
```diff
@@ -56,7 +56,13 @@ def clean_project_paths(repo_path: Path, project_paths: list[str]) -> None:
 
 def checkout_commit(repo_path: Path, commit: str) -> None:
     logger.info(f"Checking out commit: {commit}")
-    subprocess.run(["git", "checkout", commit], cwd=repo_path, stdout=subprocess.DEVNULL, stderr=subprocess.PIPE, check=True)
+    result = subprocess.run(["git", "checkout", commit], cwd=repo_path, stdout=subprocess.DEVNULL, stderr=subprocess.PIPE, check=False)
+    if result.returncode != 0:
+        # The commit may not be present locally (e.g. a real PR's base_commit that
+        # isn't in the shared clone). Fetch that SHA directly, then retry.
+        logger.info(f"Commit {commit} not present locally; fetching from origin")
+        subprocess.run(["git", "fetch", "--depth", "1", "origin", commit], cwd=repo_path, stdout=subprocess.DEVNULL, stderr=subprocess.PIPE, check=True)
```
Groenbech96, line 64: What is the harm of just running is always before the git checkout commit? Simpler logic.

Groenbech96 (approved):

src/bcbench/operations/git_operations.py
gggdttt: Always-fetching would add one redundant `git fetch origin <sha>` per entry (re-fetching what the clone step just pulled) — an extra network round-trip + auth on every matrix job, and it turns a locally-deterministic checkout into a network-dependent one (flakier under throttling). The try-first keeps the network path only for the case it's meant for: a reused testbed, or a `base_commit` outside the depth-200 window.

PR #753, opened 2026-07-28
Support plugins changes from private branchs

haoranpb:
## Why

Team needs to be able to test their changes on the plugins before merging. Currently, Copilot CLI only supports loading from the default branch (requires merging).

## How

Instead of relying on the marketplace and things like `copilot marketplace add`, we orchestrate clone the repo containing plugins ourselves and load them using `--plugin-dir`.

---

Also bundled in some other changes like moving hardcoded path into config and set the plugin dir under BC-Bench repo.


gggdttt (dismissed):

src/bcbench/operations/plugin_operations.py
```diff
@@ -1,161 +0,0 @@
-"""Install agent plugins (marketplace or local) declared in config, via the CLI's plugin commands.
```
Groenbech96, line 1: What was wrong with this?

tests/test_plugin_operations.py
```diff
@@ -1,299 +0,0 @@
-import json
-from pathlib import Path
```
Groenbech96, line 2: Same question.

src/bcbench/operations/plugin_operations.py
haoranpb: Most of the logic are refactored into `plugin.py` instead, to keep plugin related logic together

tests/test_plugin_operations.py
haoranpb: tbh, didn't look into this, left all the test related code to coding agent...

But the reason is the same, refactored into plugin.py instead of having two separate plugin related files.

Groenbech96 (dismissed):

gggdttt (approved):

PR #759, opened 2026-07-29
[Experiment] Code Review with BCQuality plugin

haoranpb:
Try experimenting with BCQuality plugin, for code review skill


haoranpb:
See #765 instead

PR #764, opened 2026-07-30
Accept root plugin.json manifest and grant plugin dir access

gggdttt:
## What

Small fixes to the shared plugin loader so a read-heavy config plugin can be consumed during an evaluation, scoped so they do not broaden the loader's contract for agents or plugins that don't need it:

1. **Accept a root `plugin.json` manifest - for Copilot only.** `_resolve_plugin` previously required the manifest at `.claude-plugin/plugin.json`. Copilot CLI also loads a `plugin.json` at the plugin root (BCQuality keeps its manifest there), while Claude Code loads only the nested layout. `resolve_config_plugins` now takes `allow_root_manifest`: the Copilot agent passes `True` (root or nested accepted); the Claude agent keeps the strict default (`False`), so a root-only plugin is **not** forwarded to a Claude run that would silently ignore it. The `AgentError` message lists the location(s) actually accepted for that run.
2. **Grant directory access only to plugins that opt in.** `run_agent` passes `--plugin-dir <dir>` for every enabled plugin, which registers/auto-discovers its skills, but Copilot CLI keeps *directory access* as a sandbox separate from `--allow-all-tools`, so a knowledge-heavy plugin registers its skill yet cannot read its own files. `--add-dir` closes that gap - but it grants **read+write**, so applying it to every plugin would over-grant and couple "plugin enabled" with "directory access granted." A new per-plugin `grant_dir_access` flag (default `false`) gates it: the Copilot agent passes `--add-dir` only for plugins that set it, mirroring how the production engine grants `--add-dir` for its knowledge root.
3. **Docstring** clarifies which manifest layout each CLI loads.

## Why

BC-Bench already loads config plugins via `resolve_config_plugins` -> `--plugin-dir` (the existing plugin mechanism). These gaps are what stop that mechanism working for a read-heavy plugin end-to-end, and the fixes are scoped so they change behavior for no other agent or plugin.

## Scope / safety

- **No shipped-config changes.** All plugins remain `enabled: false` and no plugin sets `grant_dir_access`; the `TestShippedConfig` invariants (`test_shipped_entries_are_disabled_by_default`, `test_shipped_config_resolves_on_any_os`) stay green. Enabling a plugin (and granting it dir access) for a specific run is a separate, experiment-only change.
- **Claude path stays strict.** The shared resolver keeps the nested-only contract for Claude, so this PR cannot make a root-only plugin pass validation for a Claude run.

## Tests

- `test_local_plugin_with_root_manifest_resolves_for_copilot` (root manifest resolves under `allow_root_manifest=True`) and `test_root_manifest_is_rejected_without_opt_in` (the default rejects it).
- `TestPluginsNeedingDirAccess`: only enabled, opted-in plugins receive dir access; none by default.
- Plugin suite green; `ruff check` clean.

### DIFF
--- a/src/bcbench/agent/claude/agent.py
+++ b/src/bcbench/agent/claude/agent.py
@@ -83,10 +83,9 @@ def run_claude_code(
         # --add-dir grants read+write (unlike --plugin-dir, which only registers a plugin), so hand it
         # only to plugins that opt in via grant_dir_access - currently a temporary accommodation for
         # BCQuality, whose skill reads its own knowledge files at runtime. Enabling a plugin must not
-        # silently widen the agent's sandbox access.
-        cmd_args.extend(
-            f"--add-dir={plugins[PluginConfig(**plugin_entry).record]}" for plugin_entry in claude_config["plugins"] if plugin_entry.get("enabled") and plugin_entry.get("grant_dir_access")
-        )
+        # silently widen the agent's sandbox access. Read the opt-in from the validated PluginConfig, not
+        # the raw YAML, so a non-boolean like `grant_dir_access: "false"` is not treated as truthy.
+        cmd_args.extend(f"--add-dir={plugins[plugin.record]}" for plugin in (PluginConfig(**entry) for entry in claude_config["plugins"]) if plugin.enabled and plugin.grant_dir_access)
         if custom_agent:
             cmd_args.append(f"--agent={custom_agent}")
         cmd_args.extend(
--- a/src/bcbench/agent/claude/agent.py
+++ b/src/bcbench/agent/claude/agent.py
@@ -49,7 +49,7 @@ def run_claude_code(
     skills_enabled: bool = setup_agent_skills(claude_config, entry, repo_path, agent_type=AgentType.CLAUDE)
     custom_agent: str | None = setup_custom_agent(claude_config, entry, repo_path, agent_type=AgentType.CLAUDE)
     tool_log_path: Path = setup_hooks(repo_path, AgentType.CLAUDE, output_dir)
-    plugins: list[PluginConfig] = resolve_config_plugins(claude_config, allow_copilot_manifest=False)
+    plugins: list[tuple[PluginConfig, Path]] = resolve_config_plugins(claude_config, allow_copilot_manifest=False)
 
     config = ExperimentConfiguration(
         mcp_servers=mcp_server_names,
@@ -57,7 +57,7 @@ def run_claude_code(
         custom_instructions=instructions_enabled,
         skills_enabled=skills_enabled,
         custom_agent=custom_agent,
-        plugins=[plugin.record for plugin in plugins] or None,
+        plugins=[plugin.record for plugin, _ in plugins] or None,
     )
 
     logger.info(f"Executing Claude Code in directory: {repo_path}")
@@ -79,12 +79,12 @@ def run_claude_code(
             cmd_args.append(f"--mcp-config={mcp_config_json}")
         if lsp_plugin_dir is not None:
             cmd_args.append(f"--plugin-dir={lsp_plugin_dir}")
-        cmd_args.extend(f"--plugin-dir={plugin.plugin_dir}" for plugin in plugins)
+        cmd_args.extend(f"--plugin-dir={plugin_dir}" for _, plugin_dir in plugins)
         # --add-dir grants read+write (unlike --plugin-dir, which only registers a plugin), so hand it
         # only to plugins that opt in via grant_dir_access - currently a temporary accommodation for
         # BCQuality, whose skill reads its own knowledge files at runtime. Enabling a plugin must not
         # silently widen the agent's sandbox access.
-        cmd_args.extend(f"--add-dir={plugin.plugin_dir}" for plugin in plugins if plugin.grant_dir_access)
+        cmd_args.extend(f"--add-dir={plugin_dir}" for plugin, plugin_dir in plugins if plugin.grant_dir_access)
         if custom_agent:
             cmd_args.append(f"--agent={custom_agent}")
         cmd_args.extend(
--- a/src/bcbench/agent/claude/agent.py
+++ b/src/bcbench/agent/claude/agent.py
@@ -49,7 +49,7 @@ def run_claude_code(
     skills_enabled: bool = setup_agent_skills(claude_config, entry, repo_path, agent_type=AgentType.CLAUDE)
     custom_agent: str | None = setup_custom_agent(claude_config, entry, repo_path, agent_type=AgentType.CLAUDE)
     tool_log_path: Path = setup_hooks(repo_path, AgentType.CLAUDE, output_dir)
-    plugins: dict[str, Path] = resolve_config_plugins(claude_config)
+    plugins: dict[str, Path] = resolve_config_plugins(claude_config, allow_copilot_manifest=False)
 
     config = ExperimentConfiguration(
         mcp_servers=mcp_server_names,
--- a/src/bcbench/agent/claude/agent.py
+++ b/src/bcbench/agent/claude/agent.py
@@ -83,9 +83,11 @@ def run_claude_code(
         # --add-dir grants read+write (unlike --plugin-dir, which only registers a plugin), so hand it
         # only to plugins that opt in via grant_dir_access - currently a temporary accommodation for
         # BCQuality, whose skill reads its own knowledge files at runtime. Enabling a plugin must not
-        # silently widen the agent's sandbox access. Read the opt-in from the validated PluginConfig, not
-        # the raw YAML, so a non-boolean like `grant_dir_access: "false"` is not treated as truthy.
-        cmd_args.extend(f"--add-dir={plugins[plugin.record]}" for plugin in (PluginConfig(**entry) for entry in claude_config["plugins"]) if plugin.enabled and plugin.grant_dir_access)
+        # silently widen the agent's sandbox access. Mirror resolve_config_plugins: filter on the raw
+        # `enabled` flag before constructing PluginConfig (a disabled, possibly platform-specific entry
+        # must never be validated), then gate on the validated grant_dir_access so a non-boolean like
+        # `grant_dir_access: "false"` is not treated as truthy.
+        cmd_args.extend(f"--add-dir={plugins[plugin.record]}" for plugin in (PluginConfig(**entry) for entry in claude_config["plugins"] if entry.get("enabled", False)) if plugin.grant_dir_access)
         if custom_agent:
             cmd_args.append(f"--agent={custom_agent}")
         cmd_args.extend(
--- a/src/bcbench/agent/claude/agent.py
+++ b/src/bcbench/agent/claude/agent.py
@@ -49,7 +49,7 @@ def run_claude_code(
     skills_enabled: bool = setup_agent_skills(claude_config, entry, repo_path, agent_type=AgentType.CLAUDE)
     custom_agent: str | None = setup_custom_agent(claude_config, entry, repo_path, agent_type=AgentType.CLAUDE)
     tool_log_path: Path = setup_hooks(repo_path, AgentType.CLAUDE, output_dir)
-    plugins: dict[str, Path] = resolve_config_plugins(claude_config, allow_copilot_manifest=False)
+    plugins: list[PluginConfig] = resolve_config_plugins(claude_config, allow_copilot_manifest=False)
 
     config = ExperimentConfiguration(
         mcp_servers=mcp_server_names,
@@ -57,7 +57,7 @@ def run_claude_code(
         custom_instructions=instructions_enabled,
         skills_enabled=skills_enabled,
         custom_agent=custom_agent,
-        plugins=list(plugins.keys()) or None,
+        plugins=[plugin.record for plugin in plugins] or None,
     )
 
     logger.info(f"Executing Claude Code in directory: {repo_path}")
@@ -79,15 +79,12 @@ def run_claude_code(
             cmd_args.append(f"--mcp-config={mcp_config_json}")
         if lsp_plugin_dir is not None:
             cmd_args.append(f"--plugin-dir={lsp_plugin_dir}")
-        cmd_args.extend(f"--plugin-dir={plugin_dir}" for plugin_dir in plugins.values())
+        cmd_args.extend(f"--plugin-dir={plugin.plugin_dir}" for plugin in plugins)
         # --add-dir grants read+write (unlike --plugin-dir, which only registers a plugin), so hand it
         # only to plugins that opt in via grant_dir_access - currently a temporary accommodation for
         # BCQuality, whose skill reads its own knowledge files at runtime. Enabling a plugin must not
-        # silently widen the agent's sandbox access. Mirror resolve_config_plugins: filter on the raw
-        # `enabled` flag before constructing PluginConfig (a disabled, possibly platform-specific entry
-        # must never be validated), then gate on the validated grant_dir_access so a non-boolean like
-        # `grant_dir_access: "false"` is not treated as truthy.
-        cmd_args.extend(f"--add-dir={plugins[plugin.record]}" for plugin in (PluginConfig(**entry) for entry in claude_config["plugins"] if entry.get("enabled", False)) if plugin.grant_dir_access)
+        # silently widen the agent's sandbox access.
+        cmd_args.extend(f"--add-dir={plugin.plugin_dir}" for plugin in plugins if plugin.grant_dir_access)
         if custom_agent:
             cmd_args.append(f"--agent={custom_agent}")
         cmd_args.extend(
--- a/src/bcbench/agent/claude/agent.py
+++ b/src/bcbench/agent/claude/agent.py
@@ -12,7 +12,7 @@ from bcbench.dataset import BaseDatasetEntry
 from bcbench.exceptions import AgentError, AgentTimeoutError
 from bcbench.logger import get_logger
 from bcbench.operations import setup_agent_skills, setup_custom_agent, setup_hooks, setup_instructions_from_config
-from bcbench.types import AgentMetrics, AgentType, EvaluationCategory, ExperimentConfiguration
+from bcbench.types import AgentMetrics, AgentType, EvaluationCategory, ExperimentConfiguration, PluginConfig
 
 logger = get_logger(__name__)
 _config = get_config()
@@ -80,6 +80,13 @@ def run_claude_code(
         if lsp_plugin_dir is not None:
             cmd_args.append(f"--plugin-dir={lsp_plugin_dir}")
         cmd_args.extend(f"--plugin-dir={plugin_dir}" for plugin_dir in plugins.values())
+        # --add-dir grants read+write (unlike --plugin-dir, which only registers a plugin), so hand it
+        # only to plugins that opt in via grant_dir_access - currently a temporary accommodation for
+        # BCQuality, whose skill reads its own knowledge files at runtime. Enabling a plugin must not
+        # silently widen the agent's sandbox access.
+        cmd_args.extend(
+            f"--add-dir={plugins[PluginConfig(**plugin_entry).record]}" for plugin_entry in claude_config["plugins"] if plugin_entry.get("enabled") and plugin_entry.get("grant_dir_access")
+        )
         if custom_agent:
             cmd_args.append(f"--agent={custom_agent}")
         cmd_args.extend(
--- a/src/bcbench/agent/copilot/agent.py
+++ b/src/bcbench/agent/copilot/agent.py
@@ -85,6 +85,9 @@ def run_copilot_agent(
         if lsp_plugin_dir is not None:
             cmd_args.append(f"--plugin-dir={lsp_plugin_dir}")
         cmd_args.extend(f"--plugin-dir={plugin_dir}" for plugin_dir in plugins.values())
+        # Plugins are cloned outside the review workspace, so grant the agent read access to each
+        # plugin's tree (e.g. a skill reading its own knowledge files); --plugin-dir only registers it.
+        cmd_args.extend(f"--add-dir={plugin_dir}" for plugin_dir in plugins.values())
         if custom_agent:
             cmd_args.append(f"--agent={custom_agent}")
 
--- a/src/bcbench/agent/copilot/agent.py
+++ b/src/bcbench/agent/copilot/agent.py
@@ -8,14 +8,14 @@ from pathlib import Path
 import yaml
 
 from bcbench.agent.copilot.metrics import parse_metrics
-from bcbench.agent.shared import build_al_lsp_plugin, build_mcp_config, build_prompt, parse_tool_usage_from_hooks, plugins_needing_dir_access, resolve_config_plugins
+from bcbench.agent.shared import build_al_lsp_plugin, build_mcp_config, build_prompt, parse_tool_usage_from_hooks, resolve_config_plugins
 from bcbench.config import get_config
 from bcbench.copilot_cli import find_copilot
 from bcbench.dataset import BaseDatasetEntry
 from bcbench.exceptions import AgentError, AgentTimeoutError
 from bcbench.logger import get_logger
 from bcbench.operations import setup_agent_skills, setup_custom_agent, setup_hooks, setup_instructions_from_config
-from bcbench.types import AgentMetrics, AgentType, EvaluationCategory, ExperimentConfiguration
+from bcbench.types import AgentMetrics, AgentType, EvaluationCategory, ExperimentConfiguration, PluginConfig
 
 logger = get_logger(__name__)
 _config = get_config()
@@ -54,7 +54,7 @@ def run_copilot_agent(
     skills_enabled: bool = setup_agent_skills(copilot_config, entry, repo_path, agent_type=AgentType.COPILOT)
     custom_agent: str | None = setup_custom_agent(copilot_config, entry, repo_path, agent_type=AgentType.COPILOT)
     tool_log_path: Path = setup_hooks(repo_path, AgentType.COPILOT, output_dir)
-    plugins: dict[str, Path] = resolve_config_plugins(copilot_config, allow_root_manifest=True)
+    plugins: dict[str, Path] = resolve_config_plugins(copilot_config, allow_copilot_manifest=True)
 
     config = ExperimentConfiguration(
         mcp_servers=mcp_server_names,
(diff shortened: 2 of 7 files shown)


src/bcbench/agent/shared/plugin.py
```diff
@@ -65,6 +65,16 @@ def resolve_config_plugins(agent_config: dict) -> dict[str, Path]:
     return {plugin.record: _resolve_plugin(plugin) for plugin in plugins}
 
 
+def _has_plugin_manifest(plugin_dir: Path) -> bool:
+    """True if the folder holds a plugin manifest in either supported location.
+
+    Claude Code expects ``.claude-plugin/plugin.json``; GitHub Copilot CLI also loads a
+    ``plugin.json`` at the plugin root. A plugin published in either layout is loadable.
```
haoranpb: Good change, can you be bit more explict that GitHub Copilot CLI also respects `.claude-plugin/plugin.json`?

src/bcbench/agent/copilot/agent.py
```diff
@@ -85,6 +85,9 @@ def run_copilot_agent(
         if lsp_plugin_dir is not None:
             cmd_args.append(f"--plugin-dir={lsp_plugin_dir}")
         cmd_args.extend(f"--plugin-dir={plugin_dir}" for plugin_dir in plugins.values())
+        # Plugins are cloned outside the review workspace, so grant the agent read access to each
+        # plugin's tree (e.g. a skill reading its own knowledge files); --plugin-dir only registers it.
+        cmd_args.extend(f"--add-dir={plugin_dir}" for plugin_dir in plugins.values())
         if custom_agent:
```
haoranpb: This part of the change worries me a bit. Reading through:
* https://docs.github.com/en/copilot/concepts/agents/about-plugins
* https://code.claude.com/docs/en/plugins

Realize this is not a standard plugin design => agent harness requires access to the plugin dir.

From the design of BCQuality, feels like we are not fully leveraging the power of plugions, maybe devide the content into more SKILLs, or serve them through MCP server so you could have programmatical control?

From BC-Bench itselve, whenever we experiment with BCQuality plugin, it'd mean changing two variables at once:
1. BCQuality plugin enabled
2. Granting agents access to a new dir ==> the agent will have **read and write** access, `--add-dir` is not read-only

----

I understand it takes time to change the design of BCQuality (if we were to do it), to faster move forward, maybe a new config `grant-dir-access` is needed, so we could limit the impact to BCQaulity?

src/bcbench/agent/shared/plugin.py
```diff
@@ -65,6 +65,16 @@ def resolve_config_plugins(agent_config: dict) -> dict[str, Path]:
...
+    """
+    root_manifest: Path = plugin_dir / _config.file_patterns.plugin_manifest.name
+    return (plugin_dir / _config.file_patterns.plugin_manifest).is_file() or root_manifest.is_file()
```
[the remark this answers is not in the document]
gggdttt: Good catch - fixed in cbf6a22. `resolve_config_plugins` now takes `allow_root_manifest`: the Copilot agent passes `True` while the Claude agent keeps the strict default (`False`), so a root-only manifest can no longer pass validation for a Claude run that would ignore it. Added `test_root_manifest_is_rejected_without_opt_in` to lock this in.

src/bcbench/agent/shared/plugin.py
```diff
@@ -74,8 +84,8 @@ def _resolve_plugin(plugin: PluginConfig) -> Path:
             clone_repo_at_revision(str(plugin.repo), str(plugin.revision), clone_dir)
             plugin_dir = _plugin_dir_in_clone(clone_dir, plugin)
 
-    if not (plugin_dir / _config.file_patterns.plugin_manifest).is_file():
-        raise AgentError(f"Plugin '{plugin.name}' has no manifest at {plugin_dir / _config.file_patterns.plugin_manifest}")
+    if not _has_plugin_manifest(plugin_dir):
+        raise AgentError(f"Plugin '{plugin.name}' has no manifest at {plugin_dir / _config.file_patterns.plugin_manifest} or {plugin_dir / _config.file_patterns.plugin_manifest.name}")
```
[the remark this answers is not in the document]
gggdttt: You're right, the message did change. Rather than revert it, I updated the PR description to match: the error now lists the location(s) actually accepted for the run (`{nested}` for Claude, `{nested} or {root}` for Copilot), which is the more useful behavior. (cbf6a22)

src/bcbench/agent/shared/plugin.py
gggdttt: Done in cbf6a22. The docstring now spells out that Claude Code loads only `.claude-plugin/plugin.json`, while Copilot CLI loads that layout **and** additionally a root `plugin.json` - and that the root manifest is honored only under `allow_root_manifest` (Copilot runs).

src/bcbench/agent/copilot/agent.py
gggdttt: Agreed on both points - `--add-dir` is read+write, and applying it to every plugin coupled "plugin enabled" with "directory access granted". Implemented your `grant-dir-access` suggestion as a per-plugin `grant_dir_access` flag (default `false`) on `PluginConfig`: the Copilot agent now passes `--add-dir` only for plugins that opt in (new `plugins_needing_dir_access`), so no shipped plugin gets dir access and enabling a plugin no longer widens the sandbox. Only BCQuality sets it, on the experiment branch (#765). (cbf6a22)

On the deeper point: you're right that needing raw dir read/write isn't the ideal plugin shape - the proper fix is BCQuality exposing its knowledge via skills / an MCP server instead of filesystem reads. That's a larger BCQuality-side change; `grant_dir_access` keeps the blast radius contained to BCQuality in the meantime.

src/bcbench/agent/copilot/agent.py
```diff
@@ -85,6 +85,12 @@ def run_copilot_agent(
         if lsp_plugin_dir is not None:
             cmd_args.append(f"--plugin-dir={lsp_plugin_dir}")
         cmd_args.extend(f"--plugin-dir={plugin_dir}" for plugin_dir in plugins.values())
+        # A plugin that reads its own tree at runtime (e.g. a skill loading its knowledge files)
+        # needs filesystem access beyond --plugin-dir, which only registers it. --add-dir grants
+        # read+write, so it is limited to plugins that opt in via grant_dir_access rather than
+        # handed to every plugin - keeping "plugin enabled" and "dir access granted" separate.
+        granted = plugins_needing_dir_access(copilot_config)
+        cmd_args.extend(f"--add-dir={plugins[record]}" for record in plugins if record in granted)
```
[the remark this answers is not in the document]
gggdttt: Fixed in 56335a6. You're right that `record` (`name@revision-or-source`) isn't guaranteed unique, and the collision case would silently drop one entry in the resolved dict while the survivor's path could inherit the other's `grant_dir_access` opt-in. I went with the "reject duplicate records" option: enabled-entry parsing is now centralized in `_enabled_plugins`, which raises `AgentError` when two enabled plugins share a record - so both `resolve_config_plugins` and `plugins_needing_dir_access` fail fast instead of resolving an ambiguous config. This also closes the pre-existing silent-drop, since `record` is already assumed unique as a result key. Added `test_duplicate_records_are_rejected` covering both entry points.

src/bcbench/types.py
```diff
@@ -129,6 +129,9 @@ class PluginConfig(BaseModel):
     source: PluginSource
     path: str
     enabled: bool = False
+    # Grant the agent filesystem access to this plugin's tree via Copilot `--add-dir`.
+    # Off by default so enabling a plugin does not also widen the agent's sandbox access.
+    grant_dir_access: bool = False
```
haoranpb, line 136: Add in comments that this is a (hopefully temporary) design for BCQuality

src/bcbench/agent/shared/plugin.py
```diff
@@ -44,7 +44,7 @@ def remove_agent_plugin(folder: str) -> None:
         logger.info(f"Removed stale agent plugin '{folder}': {plugin_dir}")
 
 
-def resolve_config_plugins(agent_config: dict) -> dict[str, Path]:
+def resolve_config_plugins(agent_config: dict, *, allow_root_manifest: bool = False) -> dict[str, Path]:
```
haoranpb: Good change here, maybe make it more explict that this is to support Copilot CLI.

```suggestion
def resolve_config_plugins(agent_config: dict, allow_copilot_manifest: bool = False) -> dict[str, Path]:
```

You can also shorten the docstring for it now the naming is more explicit.

src/bcbench/agent/shared/plugin.py
```diff
@@ -54,18 +54,60 @@ def resolve_config_plugins(agent_config: dict) -> dict[str, Path]:
...
+    A record (``name@revision-or-source``) keys both the resolved plugin folder and the plugin's
+    entry on a result. A collision would silently drop a plugin and, once ``grant_dir_access`` is in
+    play, let one entry's opt-in leak onto another entry's resolved path - so reject duplicates up front.
     """
     plugins = [PluginConfig(**entry) for entry in agent_config["plugins"] if entry.get("enabled", False)]
-    return {plugin.record: _resolve_plugin(plugin) for plugin in plugins}
+    records = [plugin.record for plugin in plugins]
+    duplicates = sorted({record for record in records if records.count(record) > 1})
+    if duplicates:
+        raise AgentError(f"Duplicate plugin record(s) among enabled plugins: {', '.join(duplicates)}")
```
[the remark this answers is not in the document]
haoranpb: You can probably leave the duplicate detection here, not sure if it's necessary at this stage.

When there is a need for it, we could enforce certain naming or something for the enabled plugins so this is easier.

src/bcbench/types.py
haoranpb: Also update the config.yaml for the BCQuality entry, explicitly set it to true

src/bcbench/agent/shared/plugin.py
```diff
@@ -54,18 +54,60 @@ def resolve_config_plugins(agent_config: dict) -> dict[str, Path]:
...
+
+
+def plugins_needing_dir_access(agent_config: dict) -> set[str]:
+    """Records of enabled plugins that opt into filesystem access via Copilot ``--add-dir``.
+
+    ``--add-dir`` grants read+write access, so a plugin receives it only when it sets
+    ``grant_dir_access`` in its config entry. Enabling a plugin therefore never widens the agent's
+    sandbox access on its own, keeping "plugin enabled" and "directory access granted" separate.
+    """
+    return {plugin.record for plugin in _enabled_plugins(agent_config) if plugin.grant_dir_access}
```
haoranpb: this feels a bit of overkill, in the agent.py, just a simple loop over plugins and check for `grant_dir_access` is probably enough, unlikely we will re-use this function

src/bcbench/agent/copilot/agent.py
haoranpb, line 1: Don't forget similar changes for Claude agent.y

src/bcbench/agent/copilot/agent.py
```diff
@@ -54,7 +54,7 @@ def run_copilot_agent(
     skills_enabled: bool = setup_agent_skills(copilot_config, entry, repo_path, agent_type=AgentType.COPILOT)
     custom_agent: str | None = setup_custom_agent(copilot_config, entry, repo_path, agent_type=AgentType.COPILOT)
     tool_log_path: Path = setup_hooks(repo_path, AgentType.COPILOT, output_dir)
-    plugins: dict[str, Path] = resolve_config_plugins(copilot_config)
+    plugins: dict[str, Path] = resolve_config_plugins(copilot_config, allow_root_manifest=True)
```
haoranpb: After the rename, you could change here to `allow...=(agent_type=AgentType.COPILOT)`

src/bcbench/agent/shared/plugin.py
gggdttt: Renamed `allow_root_manifest` -> `allow_copilot_manifest` (makes it explicit that a root `plugin.json` is a Copilot-only layout) and shortened the docstring now the naming carries the intent.

src/bcbench/agent/shared/plugin.py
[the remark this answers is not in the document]
gggdttt: Good catch. Added a second guard: after resolving, `resolve_config_plugins` now rejects distinct records whose resolved folders collide (two GitHub entries sharing a name clone into the same `<plugin_root>/<name>`), so a non-opted-in plugin can no longer inherit another's `--add-dir`. Covered by `test_distinct_records_resolving_to_one_folder_are_rejected`.

src/bcbench/agent/shared/plugin.py
gggdttt: Agreed. Dropped `plugins_needing_dir_access` and inlined the opt-in as a simple loop over the enabled entries in `copilot/agent.py`.

src/bcbench/agent/copilot/agent.py
gggdttt: Corrected: Claude Code's CLI also accepts `--add-dir` (per the CLI reference), so `grant_dir_access` is not Copilot-only. I mirrored the same `--add-dir` opt-in loop into `run_claude_code` (161ba5c) so a plugin that opts in gets filesystem access under both agents. Only root-manifest loading stays Copilot-specific - Claude cannot load a root `plugin.json`, so it keeps `allow_copilot_manifest=False`.

src/bcbench/agent/copilot/agent.py
gggdttt: Updated. The Copilot callsite now reads `resolve_config_plugins(copilot_config, allow_copilot_manifest=True)`. (`run_copilot_agent` hardcodes `AgentType.COPILOT` rather than taking an `agent_type` variable, so this is the literal `True`.)

src/bcbench/types.py
gggdttt: Done. Surfaced `grant_dir_access` in `config.yaml`: documented the field in the plugins header and set it explicitly on the BCQuality entry - `false` here, since BCQuality stays disabled in this PR (the stacked experiment #765 flips both `enabled` and `grant_dir_access` to `true`). Also added an inline note that granting BCQuality's skill `--add-dir` access is a (hopefully temporary) design that may change once its knowledge is served via skills / an MCP server. (acf33ea)

src/bcbench/agent/copilot/agent.py
```diff
@@ -85,6 +85,13 @@ def run_copilot_agent(
         if lsp_plugin_dir is not None:
             cmd_args.append(f"--plugin-dir={lsp_plugin_dir}")
         cmd_args.extend(f"--plugin-dir={plugin_dir}" for plugin_dir in plugins.values())
+        # --add-dir grants read+write (unlike --plugin-dir, which only registers a plugin), so hand it
+        # only to plugins that opt in via grant_dir_access - currently a temporary accommodation for
+        # BCQuality, whose skill reads its own knowledge files at runtime. Enabling a plugin must not
+        # silently widen the agent's sandbox access.
+        cmd_args.extend(
+            f"--add-dir={plugins[PluginConfig(**plugin_entry).record]}" for plugin_entry in copilot_config["plugins"] if plugin_entry.get("enabled") and plugin_entry.get("grant_dir_access")
+        )
```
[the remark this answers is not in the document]
haoranpb: See the `plugins` variable

src/bcbench/agent/claude/agent.py
```diff
@@ -80,6 +80,13 @@ def run_claude_code(
         if lsp_plugin_dir is not None:
             cmd_args.append(f"--plugin-dir={lsp_plugin_dir}")
         cmd_args.extend(f"--plugin-dir={plugin_dir}" for plugin_dir in plugins.values())
+        # --add-dir grants read+write (unlike --plugin-dir, which only registers a plugin), so hand it
+        # only to plugins that opt in via grant_dir_access - currently a temporary accommodation for
+        # BCQuality, whose skill reads its own knowledge files at runtime. Enabling a plugin must not
+        # silently widen the agent's sandbox access.
+        cmd_args.extend(
+            f"--add-dir={plugins[PluginConfig(**plugin_entry).record]}" for plugin_entry in claude_config["plugins"] if plugin_entry.get("enabled") and plugin_entry.get("grant_dir_access")
+        )
```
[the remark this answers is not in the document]
haoranpb: See `plugins` variable, you should be able to leverage it to read the config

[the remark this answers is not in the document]
haoranpb: You also want `plugin_dir`

src/bcbench/agent/copilot/agent.py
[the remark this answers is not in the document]
haoranpb: You also want `plugin_dir`

src/bcbench/agent/shared/plugin.py
```diff
@@ -54,18 +54,51 @@ def resolve_config_plugins(agent_config: dict) -> dict[str, Path]:
...
+    resolved = {plugin.record: _resolve_plugin(plugin, allow_copilot_manifest) for plugin in plugins}
+
+    # Distinct records can still resolve to one folder (e.g. two GitHub entries sharing a name clone
+    # into ``<plugin_root>/<name>``, the second clobbering the first), which would grant a non-opted-in
+    # plugin the other's ``--add-dir``. Reject shared destinations too.
+    folders = [str(path) for path in resolved.values()]
+    shared_folders = sorted({folder for folder in folders if folders.count(folder) > 1})
+    if shared_folders:
+        raise AgentError(f"Distinct plugins resolve to the same folder: {', '.join(shared_folders)}")
+    return resolved
```
haoranpb: Adding the check for duplication is a good change, but this implementation adds more complexity than our current use case.

Simply check for duplicated plugin's `name` should be good enough. The `_clone_dir` is keyed by name only, `record` is only used for reporting I believe.

What do you think?

---

THe check for unique can be as simple as:
```py
names: list[str] = [plugin.name for plugin in plugins]

if len(names) != len(set(names)):
    ....
```

gggdttt: Done in d7607ea. Replaced both guards with a single name-uniqueness check over the enabled plugins, as suggested. Since `_clone_dir` keys on `name` and `record` is `name@revision-or-source`, unique names already guarantee both unique clone dirs and unique records, and `record` is only used for reporting. Dropped the resolved-folder pass and its clone mock in the tests.

src/bcbench/agent/shared/plugin.py
```diff
@@ -54,18 +54,51 @@ def resolve_config_plugins(agent_config: dict) -> dict[str, Path]:
...
     plugins = [PluginConfig(**entry) for entry in agent_config["plugins"] if entry.get("enabled", False)]
-    return {plugin.record: _resolve_plugin(plugin) for plugin in plugins}
+
+    # A record (``name@revision-or-source``) keys both the resolved folder and the plugin's entry on a
+    # result, so a collision would silently drop a plugin - and, with grant_dir_access in play, let one
+    # entry's opt-in leak onto another's path. Reject duplicate records up front.
+    records = [plugin.record for plugin in plugins]
+    duplicate_records = sorted({record for record in records if records.count(record) > 1})
+    if duplicate_records:
+        raise AgentError(f"Duplicate plugin record(s) among enabled plugins: {', '.join(duplicate_records)}")
```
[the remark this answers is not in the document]
gggdttt: Fixed in d7607ea. The resolved-folder pass is gone; the loader now rejects duplicate plugin names before any resolution. Two GitHub entries sharing a name clone into the same `<plugin_root>/<name>` regardless of `path`, so the up-front name check closes this hole as well.

src/bcbench/agent/copilot/agent.py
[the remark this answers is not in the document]
gggdttt: Fixed in d7607ea. The `--add-dir` list is now built from the validated `PluginConfig` (`plugin.enabled and plugin.grant_dir_access`) instead of raw `.get()` on the YAML, so a non-boolean like `grant_dir_access: "false"` is no longer treated as truthy. Per @haoranpb's hints, the directory comes from the resolved `plugins[plugin.record]`.

src/bcbench/agent/claude/agent.py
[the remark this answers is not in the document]
gggdttt: Fixed in d7607ea, same as the Copilot side: the Claude `--add-dir` list now reads validated `PluginConfig` booleans and takes the directory from `plugins[plugin.record]` rather than the raw YAML.

src/bcbench/agent/claude/agent.py
```diff
@@ -80,6 +80,13 @@ def run_claude_code(
         if lsp_plugin_dir is not None:
             cmd_args.append(f"--plugin-dir={lsp_plugin_dir}")
         cmd_args.extend(f"--plugin-dir={plugin_dir}" for plugin_dir in plugins.values())
+        # --add-dir grants read+write (unlike --plugin-dir, which only registers a plugin), so hand it
+        # only to plugins that opt in via grant_dir_access - currently a temporary accommodation for
+        # BCQuality, whose skill reads its own knowledge files at runtime. Enabling a plugin must not
+        # silently widen the agent's sandbox access.
+        cmd_args.extend(
+            f"--add-dir={plugins[PluginConfig(**plugin_entry).record]}" for plugin_entry in claude_config["plugins"] if plugin_entry.get("enabled") and plugin_entry.get("grant_dir_access")
```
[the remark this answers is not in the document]
gggdttt: Keeping this block intentionally. Per @haoranpb's guidance on the earlier thread ("don't forget similar changes for the Claude agent"), the `grant_dir_access -> --add-dir` opt-in is mirrored to Claude on purpose: Claude Code's CLI supports `--add-dir`, and only root-`plugin.json` loading (`allow_copilot_manifest`) stays Copilot-only. The `PluginConfig` import is still used by this block, so it is not unused. It is now fixed the same way as Copilot (validated booleans + `plugins[plugin.record]`).

src/bcbench/agent/copilot/agent.py
```diff
@@ -85,6 +85,12 @@ def run_copilot_agent(
         if lsp_plugin_dir is not None:
             cmd_args.append(f"--plugin-dir={lsp_plugin_dir}")
         cmd_args.extend(f"--plugin-dir={plugin_dir}" for plugin_dir in plugins.values())
+        # --add-dir grants read+write (unlike --plugin-dir, which only registers a plugin), so hand it
+        # only to plugins that opt in via grant_dir_access - currently a temporary accommodation for
+        # BCQuality, whose skill reads its own knowledge files at runtime. Enabling a plugin must not
+        # silently widen the agent's sandbox access. Read the opt-in from the validated PluginConfig, not
+        # the raw YAML, so a non-boolean like `grant_dir_access: "false"` is not treated as truthy.
+        cmd_args.extend(f"--add-dir={plugins[plugin.record]}" for plugin in (PluginConfig(**entry) for entry in copilot_config["plugins"]) if plugin.enabled and plugin.grant_dir_access)
```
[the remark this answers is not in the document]
gggdttt: Good catch - real bug. Fixed in 7573ff69. The genexpr now filters on the raw `enabled` flag before constructing `PluginConfig`, mirroring `resolve_config_plugins`, so a disabled entry (like the shipped Windows-only `bcbench-example` path) is never validated on non-Windows hosts. The opt-in is still gated on the validated `grant_dir_access`. Verified against the shipped config: all three entries are disabled, so zero `PluginConfig` objects are constructed and no run fails.

src/bcbench/agent/claude/agent.py
```diff
@@ -80,6 +80,12 @@ def run_claude_code(
         if lsp_plugin_dir is not None:
             cmd_args.append(f"--plugin-dir={lsp_plugin_dir}")
         cmd_args.extend(f"--plugin-dir={plugin_dir}" for plugin_dir in plugins.values())
+        # --add-dir grants read+write (unlike --plugin-dir, which only registers a plugin), so hand it
+        # only to plugins that opt in via grant_dir_access - currently a temporary accommodation for
+        # BCQuality, whose skill reads its own knowledge files at runtime. Enabling a plugin must not
+        # silently widen the agent's sandbox access. Read the opt-in from the validated PluginConfig, not
+        # the raw YAML, so a non-boolean like `grant_dir_access: "false"` is not treated as truthy.
+        cmd_args.extend(f"--add-dir={plugins[plugin.record]}" for plugin in (PluginConfig(**entry) for entry in claude_config["plugins"]) if plugin.enabled and plugin.grant_dir_access)
```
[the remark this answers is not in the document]
gggdttt: Good catch - real bug. Fixed in 7573ff69. The genexpr now filters on the raw `enabled` flag before constructing `PluginConfig`, mirroring `resolve_config_plugins`, so a disabled entry (like the shipped Windows-only `bcbench-example` path) is never validated on non-Windows hosts. The opt-in is still gated on the validated `grant_dir_access`. Verified against the shipped config: all three entries are disabled, so zero `PluginConfig` objects are constructed and no run fails.

src/bcbench/agent/copilot/agent.py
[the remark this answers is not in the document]
gggdttt: Done in f97fc35. `resolve_config_plugins` now returns `list[PluginConfig]`, each entry carrying the `plugin_dir` it was validated into. The Copilot agent reads `plugin.plugin_dir`, `plugin.record` and `plugin.grant_dir_access` straight off the resolved objects, so the `--add-dir` line no longer re-parses the raw `config.yaml`. Enabled-first filtering and dir validation now live only in the resolver.

src/bcbench/agent/claude/agent.py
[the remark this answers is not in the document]
gggdttt: Done in f97fc35. Same change on the Claude side: `resolve_config_plugins` returns `list[PluginConfig]` with `plugin_dir` populated, and the agent reads `plugin.plugin_dir` / `plugin.grant_dir_access` directly off the resolved objects instead of re-parsing raw `config.yaml`.

src/bcbench/types.py
```diff
@@ -129,9 +129,17 @@ class PluginConfig(BaseModel):
...
+    # Off by default so enabling a plugin does not also widen the agent's sandbox access.
+    # This is a (hopefully temporary) accommodation for BCQuality, whose skill reads its own
+    # knowledge files at runtime; the long-term fix is to serve that content via skills / an MCP server.
+    grant_dir_access: bool = False
     # github only
     repo: RepoSlug | None = None
     revision: CommitSha | None = None
+    # Filesystem folder this entry resolved to, set by ``resolve_config_plugins``. None on an entry
+    # parsed straight from ``config.yaml``, which has no such key.
+    plugin_dir: Path | None = None
```
haoranpb: Unfortunately, this change here will introduce an unexpected side effect => we'll be able to specify `plugin_dir` on the `config.yaml`.

src/bcbench/agent/shared/plugin.py
```diff
@@ -44,7 +44,7 @@ def remove_agent_plugin(folder: str) -> None:
         logger.info(f"Removed stale agent plugin '{folder}': {plugin_dir}")
 
 
-def resolve_config_plugins(agent_config: dict) -> dict[str, Path]:
+def resolve_config_plugins(agent_config: dict, *, allow_copilot_manifest: bool = False) -> list[PluginConfig]:
```
haoranpb: Maybe the best way to achieve this is, not putting `plugin_dir` on `PluginConfig` class after all:

```py
def resolve_config_plugins(...) -> list[tuple[PluginConfig, Path]]:
    return [
        (plugin, _resolve_plugin(plugin, allow_copilot_manifest))
        for plugin in plugins
    ]
```

---

And then it could be used like:

```py
cmd_args.extend(f"--plugin-dir={plugin_dir}" for _, plugin_dir in plugins)
cmd_args.extend(
    f"--add-dir={plugin_dir}"
    for plugin, plugin_dir in plugins
    if plugin.grant_dir_access
)
```

haoranpb (commented):
Thanks for working through those iterations, I think we have found the right design here

src/bcbench/agent/shared/plugin.py
gggdttt: Done in 04c047a, exactly as suggested - `resolve_config_plugins` returns `list[tuple[PluginConfig, Path]]`, and both agents consume it as `for _, plugin_dir in plugins` / `for plugin, plugin_dir in plugins if plugin.grant_dir_access`. Bonus: the resolved `Path` is now non-optional, so the earlier `Path | None` is gone and the tests dropped their None-narrowing.

src/bcbench/types.py
gggdttt: Good catch, fixed in 04c047a. I dropped `plugin_dir` from `PluginConfig` entirely, so the config schema stays pure and `config.yaml` can no longer set it. `resolve_config_plugins` now returns `list[tuple[PluginConfig, Path]]`, pairing each enabled plugin with the folder it resolved to.

Groenbech96 (approved):

PR #768, opened 2026-08-03
Add neutral ignored_comments bucket to code-review scoring

gggdttt:
## What

Adds an optional **`ignored_comments`** list to each code-review gold entry (`CodeReviewEntry`). A generated comment that matches an ignored comment is treated as **neutral**: it is dropped from scoring entirely — it earns **no recall** and does **not** count against **precision**.

This gives the code-review category a third comment state alongside the two it has today:

| Bucket | Field | Effect on scoring |
| --- | --- | --- |
| Required | `expected_comments` | Must be found (drives recall); a match is correct (precision). |
| Acceptable / neutral (**new**) | `ignored_comments` | Neither required nor penalized — dropped from scoring. |
| Everything else | (omission) | Any comment not matching an expected one is a false positive (precision). |

The neutral bucket captures legitimate-but-debatable or out-of-scope findings that we do not want to force the agent to raise, but also do not want to punish it for raising (the long tail of "maintainer-judgment / accepted-noise" comments).

## How it scores

1. Expected and ignored comments are structurally matched (file + line proximity). Ignored is matched only against the generated comments **left over** after expected structural matching, so expected takes precedence.
2. Both structural buckets are validated in a **single LLM judge pass** (`judge_comment_matches_split`), then split back apart by bucket - one judge subprocess per evaluation, so no extra run-to-run nondeterminism, no second calibration target, and no stale verdict-file reuse.
3. Ignored matches are removed from the scored generated set:
   - `precision = matched / (generated - ignored)`
   - `recall` is unchanged (ignored comments never grant recall).
4. **Expected always takes precedence** — ignored only ever sees the structural leftovers, so a generated comment lands in at most one bucket.

`precision_recall(0, 0, N)` already returns precision `1.0` (the "correct silence" convention), so a positive task whose only output is an ignored comment scores precision `1.0` / recall `0.0` — it is treated as effectively silent, and macro precision keys on scorable comments so it is not rewarded as if it had commented correctly.

## Surfacing

- `ignored_comment_count` is added to `CodeReviewResult`, `CodeReviewResultSummary`, and `category_metrics`.
- Micro precision in the summary subtracts the ignored total so aggregate precision stays consistent with per-task precision.
- An **Ignored** column is added to the markdown and console comment-count tables.
- `docs/code-review.md` documents the neutral bucket and the new column.

## Backward compatibility

`ignored_comments` defaults to empty on every existing gold entry, so `ignored_count = 0` and scoring is **identical to today**. All 81 gold entries are unaffected.

## Tests / validation

- New tests cover: ignored is neutral (no recall, no precision hit), ignored does not grant recall, expected precedence, empty ignored = baseline, micro precision excludes ignored, and a positive-task-with-only-ignored macro-precision case.
- `ruff format` + `ruff check` clean, `ty` clean, full suite **678 passed / 1 skipped** (was 668 + 10 new tests).

## Open decisions (safe defaults — flagging for review)

1. **Judge-gated ignored matching** — ignored matches go through the LLM judge, mirroring expected matching, for tight semantics. Both buckets share a single judge pass (`judge_comment_matches_split`), so this adds no extra LLM call. Alternative would be structural-only.
2. **Ignored counts surfaced on dashboards** — added to summary tables + `category_metrics` for transparency. Easy to hide if we'd rather keep them internal.

Happy to adjust either based on how we want to author gold entries.


src/bcbench/evaluate/codereview.py
haoranpb, line 69: [Recommendation] Unrelated to this PR, but maybe you would like to consider to `raise` when `generated_comments` is `None`.

In that scenario, most likely something is wrong and you'd like to investigate.

src/bcbench/evaluate/codereview.py
```diff
@@ -76,17 +76,31 @@ def evaluate(self, context: EvaluationContext[CodeReviewEntry]) -> None:
...
+            # Neutralize acceptable-but-not-required findings: pass the generated comments left
+            # over after expected matching through the same structural + judge gate against the
+            # entry's ignored set, then drop those from scoring (no recall credit, no precision hit).
+            ignored_matches: list[tuple[ReviewComment, ReviewComment]] = []
+            if context.entry.ignored_comments:
+                leftover = unmatched_generated(generated_comments, validated_matches)
+                ignored_structural = match_comments(context.entry.ignored_comments, leftover)
+                ignored_matches = judge_comment_matches(
+                    ignored_structural,
+                    work_dir=context.repo_path,
```
haoranpb: [Critical] This adds a second LLM judge call per evaluation, increasing run-to-run nondeterminism and requiring additional calibration.

Please consider updating `match_comments` to account for ignored comments, then submit both expected and ignored structural matches in a single `judge_comment_matches` call. The resulting verdicts can then be separated for scoring.

If taking that approach, don't forget to update your calibration function

gggdttt: Good call - fixed in 4d357a27. Expected and ignored structural matches are now judged in a single pass: `judge_comment_matches_split` concatenates the two structural buckets, runs one `judge_verdicts` call, and splits the confirmed pairs back by bucket. So there's now exactly one judge subprocess per evaluation (the same count as before the ignored bucket existed) - no extra nondeterminism and no second calibration target. Ignored is matched against the generated comments left over after expected *structural* matching, so expected still takes precedence.

On calibration: `run_calibration`/`score_calibration` run `judge_verdicts` over hand-labeled (expected, candidate) pairs, and the judge's task ("do these two describe the same issue?") is identical regardless of which bucket a pair came from - bucketing happens after the verdict. So the single-pass change keeps `judge_verdicts` semantics unchanged and the calibration harness still measures the same thing; no calibration code change was required. If you meant seeding the calibration set with ignored-style cases, happy to add a couple - let me know.

src/bcbench/evaluate/codereview.py
gggdttt: Agreed it's worth considering, but I'd like to keep it out of this PR since it's a behavior change on a different path: today a single unparseable review output scores that one task as invalid (precision/recall 0) rather than aborting the whole run, which I think we want to preserve for batch evals. Happy to pick it up separately if we decide a parse failure should be fatal.

src/bcbench/evaluate/codereview.py
```diff
@@ -76,17 +76,31 @@ def evaluate(self, context: EvaluationContext[CodeReviewEntry]) -> None:
...
+                )
```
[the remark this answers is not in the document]
gggdttt: Resolved in 4d357a27 - the second judge pass is gone. Expected and ignored pairs are now judged in a single `judge_verdicts` call (`judge_comment_matches_split`), so only one judge subprocess runs per evaluation and there's no stale `judge_results.json` to reuse.

src/bcbench/results/codereview.py
```diff
@@ -266,9 +298,9 @@ def render_github_metrics_markdown(self) -> str:
         return (
             "## Comment counts\n"
             "\n"
-            "| Generated | Expected | Matched | Incorrect | Missed |\n"
-            "|----------:|---------:|--------:|----------:|-------:|\n"
-            f"| {self.generated_comment_count} | {self.expected_comment_count} | {self.matched_comment_count} | {self.incorrect_comment_count} | {self.missed_comment_count} |\n"
+            "| Generated | Expected | Matched | Incorrect | Missed | Ignored |\n"
+            "|----------:|---------:|--------:|----------:|-------:|--------:|\n"
+            f"| {self.generated_comment_count} | {self.expected_comment_count} | {self.matched_comment_count} | {self.incorrect_comment_count} | {self.missed_comment_count} | {self.ignored_comment_count} |\n"
```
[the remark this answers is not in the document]
gggdttt: Fixed in 4d357a27. `_METRIC_EXPLANATIONS` and `_CONSOLE_METRIC_EXPLANATIONS` now define Micro as summing matched, scorable generated (generated minus ignored), and expected, and add an Ignored-comment entry stating that ignored comments are removed from the scored generated denominator.

docs/code-review.md
```diff
@@ -115,6 +117,7 @@ Compares review-knowledge configurations for the same model (see the Baseline Le
 - **Fβ (β=0.5)** — precision-leaning F-score; use when false positives are costly (noisy reviews waste reviewer time).
 - **Fβ (β=2)** — recall-leaning F-score; weights catching issues more than avoiding noise.
 - **Severity MAE** — mean absolute error between the agent's and the expected severity levels, over matched comments only. Lower is better; `0` means every matched comment got the severity exactly right.
+- **Ignored** — generated comments that matched an entry's `ignored_comments` set. These are excluded from precision (they are neither correct nor incorrect); the count is surfaced for transparency only.
```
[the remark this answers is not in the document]
gggdttt: Fixed in 4d357a27. `docs/code-review.md` now defines Precision over scorable generated comments (generated minus ignored) and updates the Micro description to sum matched, scorable generated, and expected - matching the implementation.

src/bcbench/evaluate/codereview.py
```diff
@@ -72,8 +72,15 @@ def evaluate(self, context: EvaluationContext[CodeReviewEntry]) -> None:
...
                 generated_comments,
             )
-            validated_matches = judge_comment_matches(
+            # Ignored comments are neutral: match them against the generated comments left over
+            # after expected structural matching (expected takes precedence), then judge both
+            # buckets in a SINGLE pass. One judge call per evaluation avoids a second LLM round
+            # (less nondeterminism, one calibration target) and cannot reuse a stale verdict file.
+            leftover = unmatched_generated(generated_comments, structural_matches)
+            ignored_structural = match_comments(context.entry.ignored_comments, leftover)
+            validated_matches, ignored_matches = judge_comment_matches_split(
```
haoranpb: [nit] Consider making the naming here more explicity, both variables and functions. Bit hard to tell which functions have LLM call behind the scenes, and which ones are programmatic.

e.g instead `leftover`, `unmatched_generated_comments`

src/bcbench/evaluate/codereview_judge.py
```diff
@@ -127,6 +127,29 @@ def judge_comment_matches(
     return [pair for pair, is_match in zip(matched_pairs, verdicts, strict=True) if is_match]
 
 
+def judge_comment_matches_split(
```
haoranpb: [recommandation] Replace the previous `judge_comment_matches`? If we don't plan to use the old function in the immediate future, probably easier to move than to maintain 2 similar functions.

[nit] new function name? e.g. `judge_comment_match`?

src/bcbench/results/codereview.py
```diff
@@ -130,11 +146,13 @@ class CodeReviewResult(BaseEvaluationResult):
 
     generated_comments: list[ReviewComment] = Field(default_factory=list)
     expected_comments: list[ReviewComment] = Field(default_factory=list)
+    ignored_comments: list[ReviewComment] = Field(default_factory=list)
     valid_review_output: bool = False
 
     matched_comment_count: int = Field(default=0, ge=0)
     missed_comment_count: int = Field(default=0, ge=0)
     incorrect_comment_count: int = Field(default=0, ge=0)
+    ignored_comment_count: int = Field(default=0, ge=0)
```
haoranpb: Not sure this is needed here since it's something part of the dataset.

Probably more interesting to know how many ignored comments are matched?

src/bcbench/results/codereview.py
```diff
@@ -113,6 +115,20 @@ def match_comments(
     ]
 
 
+def unmatched_generated(
```
haoranpb, line 118: [Critical] can't find any tests for this function, while it is critical in the evaluation process here

src/bcbench/results/codereview.py
```diff
@@ -151,21 +169,35 @@ def create(
...
+        ignored_comments = ignored_comments or []
+        if ignored_matched_pairs is None:
+            leftover = unmatched_generated(generated_comments, matched_pairs)
+            ignored_matched_pairs = match_comments(ignored_comments, leftover)
+        ignored_count = len(ignored_matched_pairs)
+
+        # Ignored comments are neutral: drop them from the scored generated set so they neither
+        # earn recall (they are not expected) nor cost precision (they are not false positives).
+        scored_generated_count = len(generated_comments) - ignored_count
+        precision, recall = precision_recall(matched_count, scored_generated_count, len(expected_comments))
```
haoranpb: [Recommendation] it's probably valuable to extract those code into a function so you can thoroghly test the logic.

haoranpb (changes requested):
Looks good, would be great to get test coverage on your critical metrics calculation path

src/bcbench/results/codereview.py
gggdttt: Agreed it was misleading -- it never stored the dataset size. It already counts how many ignored comments were *matched* (`len(ignored_matched_pairs)`), which is exactly the 'how many ignored got matched' number you're after. Renamed the field to `ignored_matched_comment_count` to say so.

src/bcbench/results/codereview.py
gggdttt: Done -- extracted the scoring math into a pure `_score_counts(matched, generated, expected, ignored)` helper and added `TestScoreCounts` covering the normal split, the all-ignored case (precision stays 1.0, recall 0.0), and the ignored-reduces-scored-set case.

src/bcbench/evaluate/codereview.py
gggdttt: Good call. Renamed `leftover` -> `unmatched_generated_comments`. To make the LLM boundary obvious I also renamed the split judge entry point to `judge_expected_and_ignored` (the two functions with a subprocess behind them are now `judge_verdicts` and `judge_expected_and_ignored`; everything else -- `match_comments`, `unmatched_generated`, `_score_counts` -- is purely programmatic).

src/bcbench/evaluate/codereview_judge.py
gggdttt: Done -- removed `judge_comment_matches` (it had no callers left once the pipeline moved to the single-pass path) and renamed `judge_comment_matches_split` -> `judge_expected_and_ignored`. Its mechanics tests were retargeted to `judge_verdicts`, which is the actual subprocess runner.

src/bcbench/results/codereview.py
gggdttt: Fixed -- added `TestUnmatchedGenerated` (4 cases), including one that pins down the identity-not-value-equality semantics (two value-equal `ReviewComment` instances are treated as distinct, so a matched generated comment doesn't accidentally filter out an unrelated equal one).

src/bcbench/evaluate/codereview.py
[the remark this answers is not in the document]
gggdttt: Follow-up in this round: I also tightened the same path. Ignored comments are now matched against *all* generated comments rather than only the ones left over after expected structural matching, and expected precedence is applied *after* the judge (an ignored pair whose generated comment the judge confirmed as expected is dropped). This closes a gap where a finding whose expected pair the judge rejected was never offered to the ignored bucket and wrongly counted as a false positive. Still a single `judge_verdicts` pass, so the stale-file concern stays resolved.

gggdttt:
Thanks @haoranpb - round 2 pushed (f156cbea). Summary:

**Test coverage (the blocking asks)**
- `unmatched_generated` now has `TestUnmatchedGenerated` (incl. identity-not-value-equality).
- Extracted the scoring math into a pure `_score_counts` helper with `TestScoreCounts` (normal split, all-ignored, ignored-reduces-scored-set).
- Added precedence tests to `TestJudgeExpectedAndIgnored`.

**Naming / dedup**
- `leftover` -> `unmatched_generated_comments`; `judge_comment_matches` removed; `judge_comment_matches_split` -> `judge_expected_and_ignored`. The only functions with an LLM subprocess are now `judge_verdicts` and `judge_expected_and_ignored`.
- `ignored_comment_count` -> `ignored_matched_comment_count` (it always counted matched ignored comments, not dataset size).

**Correctness (from the bot thread)**
- Ignored comments are matched against all generated comments and expected precedence is applied after the judge, so a finding whose expected pair the judge rejects can still be neutralized as ignored instead of counting as a false positive. Single judge pass preserved.

Gauntlet green: ruff format/check + ty clean, 682 passed / 1 skipped (+10 new tests).

haoranpb (approved):
Looks good overall.

One design question: the distinction between `ignored_comments` and severity is not immediately clear to me. My understanding is that severity represents impact, while `ignored_comments` represents whether a finding is required for benchmark scoring. A low-severity expected comment still affects recall, whereas an ignored comment is neutral. Could we **document that rationale and the criteria for choosing one bucket over the other**?

This also highlighted that the **core F score weights every expected finding equally**: missing a critical issue has the same effect as missing a low-severity issue. Is that intentional? If impact should affect the leaderboard, severity-weighted precision/recall could provide a more general model, potentially `ignored_comments` can simplify be zero-weight findings => much easier.

haoranpb:
> Thanks Sun Haoran (@haoranpb) - round 2 pushed ([f156cbe](https://github.com/microsoft/BC-Bench/commit/f156cbea4b8909ed1d287b74feac07f4eb399d79)). Summary:
> 
> **Test coverage (the blocking asks)**
> 
> * `unmatched_generated` now has `TestUnmatchedGenerated` (incl. identity-not-value-equality).
> * Extracted the scoring math into a pure `_score_counts` helper with `TestScoreCounts` (normal split, all-ignored, ignored-reduces-scored-set).
> * Added precedence tests to `TestJudgeExpectedAndIgnored`.
> 
> **Naming / dedup**
> 
> * `leftover` -> `unmatched_generated_comments`; `judge_comment_matches` removed; `judge_comment_matches_split` -> `judge_expected_and_ignored`. The only functions with an LLM subprocess are now `judge_verdicts` and `judge_expected_and_ignored`.
> * `ignored_comment_count` -> `ignored_matched_comment_count` (it always counted matched ignored comments, not dataset size).
> 
> **Correctness (from the bot thread)**
> 
> * Ignored comments are matched against all generated comments and expected precedence is applied after the judge, so a finding whose expected pair the judge rejects can still be neutralized as ignored instead of counting as a false positive. Single judge pass preserved.
> 
> Gauntlet green: ruff format/check + ty clean, 682 passed / 1 skipped (+10 new tests).

Good work, see above comment for the challenge on the design here.  @gggdttt

PR #770, opened 2026-08-04
Collect code-review entries per reviewed commit (bug 643990)

gggdttt:
## Why

The `codereview` collector built **one entry per PR**: `base_commit` = merge-base, `patch` = the full final PR diff (`gh pr diff`), and every inline comment re-anchored to its **head** `line`. A comment a reviewer wrote on an *earlier* commit — against a smaller diff — was then scored against the larger final patch, i.e. the agent is graded on a patch the reviewer never saw. This is the per-commit vs per-PR mismatch tracked in **bug 643990**; the team agreed on the **per-commit** contract.

## What

Each inline comment is now grouped by the commit it was written on and scored against that commit's diff:

- **Anchor to the reviewed commit.** Comments are grouped by `original_commit_id` (fallback `commit_id`, then the PR head) and anchored to `original_line` / `original_start_line` — the position in the diff GitHub rendered at review time — instead of the live head `line`.
- **Store the diff the reviewer saw.** The entry for a commit stores the three-dot range `base...original_commit_id`, reconstructed via a new `GHClient.get_compare_diff` (compare API, `Accept: application/vnd.github.diff`). For the PR head this is exactly `gh pr diff`, so that path is unchanged.
- **One entry per reviewed commit.** N distinct reviewed commits → N entries, `instance_id` suffixed `-1`, `-2`, … A single-pass PR (the common case) still yields exactly one entry with the stable `<repo>-<pr>` id, byte-identical to the old output when the pass is the PR head.
- **Empty fallback preserved.** When no comment is selected (e.g. all thumbs-down, or a `--reviewer` that matches no one) a single per-PR entry with the full diff and `expected_comments=[]` is emitted, as before.
- `collect_codereview_entry` now returns a **list** of entries; the `bcbench collect codereview` CLI reports the entry and comment counts.

WIP/no-comment intermediate commits produce no entries because entries are built only from commits that carry a selected comment.

## Tests

- Refactored the shared reviewer/reacted/placeable filter into `_select_comment`; `build_expected_comments` (per-PR/head positions) is retained and unchanged.
- Added `group_expected_by_commit` unit tests (grouping order, `original_line` anchoring, `commit_id`/fallback resolution, reply/left-side skipping) and orchestration tests (per-commit split into `-1`/`-2` using `get_compare_diff`; empty-selection per-PR fallback).
- Gauntlet green: `ruff format` (no changes) · `ruff check` · `ty check` · `pytest` = **674 passed, 1 skipped**.

Resolves bug 643990.


src/bcbench/collection/collect_codereview.py
```diff
@@ -175,28 +265,67 @@ def _build_codereview_entry(
...
+        "Built %d code-review entr%s from %d comment(s) for PR #%s",
+        len(entries),
+        "y" if len(entries) == 1 else "ies",
+        len(comments),
+        pr_number,
     )
+    return entries
 
 
 def collect_codereview_entry(
```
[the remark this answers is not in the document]
gggdttt: Good catch - renamed to collect_codereview_entries and updated the export, CLI caller, and tests in bdd98950.

gggdttt:
Addressed the suppressed review note on get_compare_diff: added TestGHClient.test_get_compare_diff_success (commit d477d61b) that stubs subprocess.run and asserts the /compare/{base}...{head} endpoint, the 'Accept: application/vnd.github.diff' header, and raw stdout passthrough. Full suite green (675 passed, 1 skipped).

src/bcbench/collection/collect_codereview.py
```diff
@@ -11,7 +20,9 @@
   guards by omission (the construct stays in the patch but is absent from
   ``expected_comments``, so flagging it costs precision).
 
-If neither is given, every top-level inline comment on the PR is used.
+If neither is given, every top-level inline comment on the PR is used. If no comment
+is selected at all, a single per-PR entry carrying the full final diff and no expected
```
haoranpb: [nit] What is "selected"?

src/bcbench/collection/collect_codereview.py
```diff
@@ -81,90 +92,169 @@ def parse_domain_severity(body: str) -> tuple[str | None, Severity | None]:
...
 
 
-def _comment_line_span(comment: dict[str, Any]) -> tuple[int, int | None] | None:
-    """Return (line_start, line_end) for a RIGHT-side comment, or None if unplaceable."""
+def _comment_line_span(comment: dict[str, Any], *, use_original: bool = False) -> tuple[int, int | None] | None:
+    """Return (line_start, line_end) for a RIGHT-side comment, or None if unplaceable.
+
+    With ``use_original`` the span is read from ``original_line`` / ``original_start_line``
+    (the comment's position in the diff of the commit it was written on) instead of the
+    live ``line`` / ``start_line`` (its position in the current head diff).
```
haoranpb: [recommendation] The variable `use_original` here is for backward-compatiblity only, not sure if it would be valuable here now that we moved on to the new approach.

I would imagine we'll only collect dataset using the new approach going forward.

haoranpb: Same for other functions

gggdttt: Good call - `use_original` and the whole per-PR `build_expected_comments` path were only kept for backward compatibility and are dead now that collection is per-commit (the orchestration only goes through `group_expected_by_commit`). Removed both in b101b5b8: `_comment_line_span`/`_select_comment` always anchor to `original_line`, and the per-PR selector is gone. Its unique filter coverage (reviewer / reacted / multiline span / path-with-spaces / unanchored skip) moved onto `group_expected_by_commit`.

src/bcbench/collection/collect_codereview.py
gggdttt: Clarified in the module docstring (b101b5b8): "selected" means a comment that survives the reviewer/reacted filters and is placeable on the diff, so it becomes an expected finding.

haoranpb (approved):

PR #776, opened 2026-08-07
Add per-article BCQuality coverage for code-review dataset (AB#646153)

gggdttt:
## What

Adds **per-article BCQuality coverage tracking** to the `code-review` category so we can see, per BCQuality knowledge article, which articles have gold coverage and which have none. Follow-up from the 2026-08-06 PR Review Agent sync (AB#646153).

## Why

Today gold entries are classified only by coarse `domain`/`area`. We cannot tell *which specific BCQuality article* an entry exercises, nor which articles have zero gold coverage. This blocks Wael's self-improving-agent pipeline from targeting coverage gaps.

## Changes

- **Schema (optional, backward-compatible)** in `dataset/codereview.py`:
  - `ArticleId` — shared `<domain>/<slug>` constrained alias (e.g. `security/secrettext-for-credentials`).
  - `ReviewComment.article` — the article a *required finding* derives from.
  - `CodeReviewEntryMetadata.articles` — entry-level list for false-positive-guard entries (`expected_comments = []`) that test an article by omission and so have no per-comment `article`. Category-specific metadata lives on this `EntryMetadata` subclass rather than the shared base.
  - A model validator keeps the two annotation sources **disjoint**, and `declared_articles()` returns their union.
  - Existing entries without annotations load unchanged (counted as *unannotated*).
- **`analysis/bcquality_article_coverage.py`** — pure analysis logic: aggregate declared articles, enumerate the BCQuality article inventory from a checkout, compute covered / zero-coverage / unknown-slug / unannotated. Kept out of `dataset`, which stays the schema/model layer.
- **`notebooks/code-review-coverage.ipynb`** — the coverage report surface, mirroring the existing `notebooks/dataset.ipynb`. Inventory comes from `$BCQUALITY_ROOT`; when absent it degrades to declared-only (zero-coverage undetermined). Rendered outputs are committed so the analysis is readable directly on GitHub.
- **Backfill:** the **security** domain is annotated as a worked example (17 of its 20 entries → 9 distinct articles). Remaining domains are intentionally left for follow-up backfill; the report surfaces them as unannotated.

## Example output (committed in the notebook)

```
9 articles covered across 17/130 annotated entries; 241 of 250 articles have zero coverage
```

| Domain | Covered | Gold entries | Inventory | Zero-cov |
| --- | --- | --- | --- | --- |
| security | 9 | 17 | 24 | 15 |
| performance | 0 | 0 | 47 | 47 |
| style | 0 | 0 | 35 | 35 |
| ui | 0 | 0 | 29 | 29 |
| ... | | | | |

The low overall ratio is expected at this stage: this PR delivers the mechanism plus a single worked domain, so 8 of 9 dataset areas are still unannotated by design.

## Validation

`ruff format`/`check` clean (including the notebook), `ty` clean (the one remaining diagnostic is a pre-existing `unused-ignore` in `redteam.py` from main, untouched here), `pytest` **738 passed / 2 skipped**. Dataset still loads all 130 entries; the JSONL diff touches only the security lines. Notebook executed end-to-end against BCQuality `main`, with non-deterministic execution timestamps stripped so the committed outputs stay reviewable.

## Notes

- Additive — no existing scoring/behavior changes.
- No CLI subcommand: this is ad-hoc analysis, so it lives in `notebooks/` per review feedback.
- Ties into the neutral `ignored_comments` work (#768): FP-guard gold cases can be classified by the article whose false positive they guard.

AB#646153


haoranpb:
@gggdttt Quick question, is this function for ad-hoc analysis only? Or would you need it for downstream consumption?

src/bcbench/dataset/dataset_entry.py
```diff
@@ -30,6 +30,10 @@ class EntryMetadata(BaseModel):
     area: str | None = None
     image_count: Annotated[int, Field(ge=0)] | None = None
     persona: str | None = None
+    # BCQuality knowledge articles this entry exercises as `<domain>/<slug>`. Primarily
+    # for false-positive-guard entries (expected_comments=[]) that test an article by
+    # omission and thus have no per-comment `article` to carry the association.
+    articles: list[Annotated[str, Field(pattern=r"^[a-z0-9][a-z0-9-]*/[a-z0-9][a-z0-9-]*$")]] = Field(default_factory=list)
```
haoranpb: The existing property `persona` here probably confused you, it is not used anywhere now (replaced by `audience` property on NL2AL dataset class). ~~It'd be nice if you can help remove it.~~ I will clean it up in the future.

[design] Since `articles` only applies to code-review entries, could we introduce `CodeReviewEntryMetadata(EntryMetadata)` subclass to include that property instead? So category-specific properties are own by their own class.

Then in the `CodeReviewEntry` class, you can probably override the metadata properties type with something like:

```python
class CodeReviewEntry(RepoGroundedEntry):
    """Dataset entry for the code-review category."""
    metadata: CodeReviewEntryMetadata = Field(default_factory=CodeReviewEntryMetadata)

```

src/bcbench/dataset/codereview.py
```diff
@@ -93,3 +97,14 @@ def get_task(self) -> str:
 
     def get_expected_output(self) -> str:
         return "\n".join(str(c) for c in self.expected_comments)
+
+    def declared_articles(self) -> set[str]:
```
haoranpb: [design] I expected `CodeReviewEntry.metadata.articles` to always be the same as `{entry.article for entry in CodeReviewEntry.expected_comments if entry.article}`, but looks like it is not the case?

And to avoid drifting between the two when you update the dataset, you can leverage the `@model_validator(mode="after")` from pydantic to assert this behavior

src/bcbench/dataset/dataset_entry.py
haoranpb: [recommendation] consider defined a shared `ArticleId` type and use it for both here and on `ReviewComment`.

src/bcbench/commands/dataset.py
```diff
@@ -182,6 +182,92 @@ def version(
...
+@dataset_app.command("coverage")
+def coverage(
+    category: EvaluationCategoryOption = EvaluationCategory.CODE_REVIEW,
+    bcquality_root: Annotated[
+        str | None,
+        typer.Option(help="Path to a BCQuality checkout; enables zero-coverage detection. Falls back to $BCQUALITY_ROOT."),
+    ] = None,
+    show_zero: Annotated[bool, typer.Option(help="List every zero-coverage article (can be long)")] = False,
+    github_output: Annotated[str | None, typer.Option(help="Write the coverage report as JSON to GITHUB_OUTPUT with this key name")] = None,
+) -> None:
```
haoranpb: This command here is very category-specific, and apearts to be purely ad-hoc analysis for now.

Unless we have immediate downstream usage, I'd prefer we move things into `notebooks`. See the pre-existing notebooks where we conduted similiar dataset analysis (for bug-fixing): [notebooks/dataset.ipynb](https://github.com/microsoft/BC-Bench/blob/main/notebooks/dataset.ipynb).


Another benefit is you can share the results/information directly, because notebook's result are rendered directly on GitHub.

src/bcbench/dataset/dataset_entry.py
gggdttt: Done. Moved rticles onto a CodeReviewEntryMetadata(EntryMetadata) subclass and overrode CodeReviewEntry.metadata to that type exactly as suggested. Left persona alone since you're cleaning it up separately.

src/bcbench/dataset/codereview.py
gggdttt: Added a `@model_validator(mode="after")` to lock the two annotation sources together, but with a disjointness invariant rather than equality -- the literal `metadata.articles == {c.article for c in expected_comments if c.article}` does not hold by design.

The two sources are complementary, not redundant:
- A finding that has an expected comment carries its article on the comment (`ReviewComment.article`); those entries keep `metadata.articles` empty.
- A false-positive-guard entry has `expected_comments == []` (it tests an article by omission), so there is no comment to carry the association -- `metadata.articles` exists precisely for that case.

So requiring equality would reject every FP-guard entry (empty comments but non-empty `metadata.articles`) and every commented entry (vice versa). Instead the validator asserts the two sets are **disjoint**: an article must not be declared in both places, which is what actually prevents drift here. `declared_articles()` still returns the union of both, so coverage counts every article regardless of which side annotates it.

src/bcbench/dataset/dataset_entry.py
gggdttt: Done. Added a shared ArticleId = Annotated[str, Field(pattern=...)] alias in codereview.py and used it for both ReviewComment.article and CodeReviewEntryMetadata.articles.

src/bcbench/commands/dataset.py
gggdttt: Agreed -- removed the CLI command and moved the analysis into 
otebooks/code-review-coverage.ipynb, matching the bug-fix 
otebooks/dataset.ipynb precedent. coverage.py stays as an importable library the notebook builds on (and for any future re-wiring). One trade-off: the notebook drops the --github-output JSON export the command had, but there is no CI consumer of it today.

dataset/codereview.jsonl
```diff
@@ -1,18 +1,18 @@
...
...dy to the user, leaking internal service details. Log the details and show a generic message."}, {"file": "src/OutlookAddinDeployer.Codeunit.al", "line_start": 30, "line_end": 30, "domain": "security", "severity": "medium", "body": "GetLastErrorText() is shown to the user, exposing internal system details. Log the raw error and present a sanitized message."}, {"file": "src/OutlookAddinDeployer.Codeunit.al", "line_start": 33, "line_end": 33, "domain": "security", "severity": "high", "body": "The access token is returned as plain Text instead of SecretText, exposing it in memory and to the debugger. Return and handle it as SecretText.", "article": "security/secrettext-for-credentials"}, {"file": "src/OutlookAddinDeployer.Codeunit.al", "line_start": 35, "line_end": 35, "domain": "security", "severity": "high", "body": "A hardcoded access token is embedded in source code. Retrieve the token from a secure store or OAuth flow instead of hardcoding it.", "article": "security/isolatedstorage-setencrypted-for-sensitive-values"}], "category": "code-review", "description": "True positive security findings: error_exposure (verified line numbers)", "expect_findings": true, "source": "vsoadmin"}
```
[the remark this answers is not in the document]
gggdttt: Same remap applied to the hardcoded-token finding here -> \security/secrettext-for-credentials\. Fixed in b1f42e61.

dataset/codereview.jsonl
```diff
@@ -1,18 +1,18 @@
...
...rdcodedSecretClient.Codeunit.al\nnew file mode 100644\n--- /dev/null\n+++ b/src/HardcodedSecretClient.Codeunit.al\n@@ -0,0 +1,18 @@\n+codeunit 50301 \"Hardcoded Secret Client\"\n+{\n+  procedure CallApi()\n+  var\n+    HttpClient: HttpClient;\n+    HttpHeaders: HttpHeaders;\n+    HttpResponseMessage: HttpResponseMessage;\n+  begin\n+    HttpHeaders := HttpClient.DefaultRequestHeaders();\n+    HttpHeaders.Add('X-Api-Key', this.GetApiKey());\n+    HttpClient.Get('https://api.contoso.com/data', HttpResponseMessage);\n+  end;\n+\n+  local procedure GetApiKey(): Text\n+  begin\n+    exit('sk-1234567890abcdef');\n+  end;\n+}\n", "expected_comments": [{"file": "src/HardcodedSecretClient.Codeunit.al", "line_start": 16, "line_end": 16, "severity": "critical", "domain": "security", "body": "Hardcoded API key in source code. Retrieve secrets from encrypted isolated storage or another secure store instead.", "article": "security/isolatedstorage-setencrypted-for-sensitive-values"}], "category": "code-review", "description": "Clean codeunit using SecretText, NonDebuggable, IsolatedStorage.SetEncrypted, and HTTPS enforcement with no security issues", "expect_findings": true, "source": "vsoadmin"}
```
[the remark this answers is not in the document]
gggdttt: Good catch -- remapped to security/secrettext-for-credentials. That article explicitly covers secrets embedded in source (a string literal cannot be assigned directly to a SecretText, which prevents embedding a secret in source code), whereas isolatedstorage-setencrypted-for-sensitive-values is scoped to the Set-vs-SetEncrypted write choice. Fixed in b1f42e61.

dataset/codereview.jsonl
```diff
@@ -1,18 +1,18 @@
...
...seAgentConsumption.Table.al", "line_start": 42, "line_end": 42, "domain": "security", "severity": "medium", "body": "The procedure accepts an arbitrary UserSecurityId, letting a caller log consumption against any user's identity. Derive the user from UserSecurityId() instead of trusting the parameter."}, {"file": "src/ExpenseAgentAdmin.PermissionSet.al", "line_start": 13, "line_end": 13, "domain": "security", "severity": "medium", "body": "RIMD on Agent Creation Control lets assigned users delete creation-control records, removing a security guardrail. Grant only the permissions actually required.", "article": "security/inherent-permissions-minimal-grant"}, {"file": "src/ExpenseAgentAdmin.PermissionSet.al", "line_start": 14, "line_end": 14, "domain": "security", "severity": "medium", "body": "IMD on Expense Report Rule Violation lets users delete recorded policy violations, enabling them to hide their own violations. Remove delete and modify access.", "article": "security/inherent-permissions-minimal-grant"}], "category": "code-review", "description": "True positive security findings: permission (trimmed to 5 representative findings)", "expect_findings": true, "source": "vsoadmin"}
```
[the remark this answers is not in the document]
gggdttt: Partially. security-006's tag is correct and stays: its finding is a genuine method-level [InherentPermissions(PermissionObjectType::TableData, Database::Customer, 'RIMD')] attribute on a procedure -- exactly what this article covers. But you're right about security-013: those findings are object-level permissionset InherentPermissions/InherentEntitlements = RIX properties and specific-table RIMD/IMD grant lines, which the method-level-attribute article does not cover. No other BCQuality article faithfully covers object-level inherent properties or non-wildcard grant breadth (permission-set-avoid-wildcard-grants is wildcard-specific), so I dropped the three tags and left those findings unannotated. Fixed in b1f42e61.

src/bcbench/dataset/codereview.py
```diff
@@ -88,8 +106,38 @@ class CodeReviewEntry(RepoGroundedEntry):
...
+        return self
+
     def get_task(self) -> str:
         return self.patch
 
     def get_expected_output(self) -> str:
         return "\n".join(str(c) for c in self.expected_comments)
+
+    def declared_articles(self) -> set[str]:
+        """BCQuality articles this entry is annotated against.
```
haoranpb, line 135: The type can be strengthened here

```suggestion

    def declared_articles(self) -> set[ArticleId]:
        """BCQuality articles this entry is annotated against.
```

src/bcbench/dataset/coverage.py
```diff
@@ -0,0 +1,138 @@
...
+_KNOWLEDGE_SUBDIR = Path("microsoft") / "knowledge"
+_BCQUALITY_ROOT_ENV = "BCQUALITY_ROOT"
+
+
+class ArticleCoverage(BaseModel):
+    """One article and the gold entries that exercise it."""
+
+    model_config = ConfigDict(frozen=True)
+
+    article: str
```
haoranpb: Article could have a stronger type
```suggestion
    article: ArticleId
```

src/bcbench/dataset/coverage.py
```diff
@@ -0,0 +1,138 @@
...
+        if not self.inventory_available:
+            return 0
+        return len(self.covered) + len(self.zero_coverage)
+
+    @property
+    def annotated_entries(self) -> int:
+        return self.total_entries - len(self.unannotated_entry_ids)
+
+
+def _domain_of(article: str) -> str:
```
haoranpb:
```suggestion
def _domain_of(article: ArticleId) -> str:
```

src/bcbench/analysis/bcquality_article_coverage.py
haoranpb, line 1: [design] `dataset` module is currently the schema/model layer. Majority of the code in this file doesn't really belong here as they are mostly for analysis.

Could we move the BCQuality specific reporting logic into a new module `bcbench.analysis.bcquality_artical_coverage` while keeping the article schema and model behavior in `dataset.codereview`?

`coverage.py` should probably be renamed to `bcquality_artical_coverage.py` so it's more clear what it is about.

haoranpb (changes requested):
Also noticed the notebook has no commited outputs, you can probably include that as well.

src/bcbench/dataset/codereview.py
gggdttt: Strengthened `CodeReviewEntry.declared_articles()` to return `set[ArticleId]` in b0a0e915.

src/bcbench/dataset/coverage.py
gggdttt: Strengthened `ArticleCoverage.article` to `ArticleId` in b0a0e915. The model now lives in `bcbench.analysis.bcquality_article_coverage` as part of the requested module move.

src/bcbench/dataset/coverage.py
gggdttt: Updated `_domain_of` to accept `ArticleId` in b0a0e915. I also propagated `ArticleId` through the coverage report collections and inventory helper types.

src/bcbench/analysis/bcquality_article_coverage.py
gggdttt: Moved all BCQuality-specific coverage reporting and inventory logic from `bcbench.dataset` to the correctly spelled `bcbench.analysis.bcquality_article_coverage` module in b0a0e915. `dataset.codereview` now contains only the article schema/model behavior, `bcbench.dataset` no longer exports analysis helpers, and the notebook/tests import the new analysis module.

gggdttt:
Committed the executed notebook outputs in b0a0e915. The rendered report now shows 9 covered articles across 17 of 130 annotated entries, a 186-article BCQuality inventory with 177 zero-coverage articles, per-domain totals, covered article-to-entry mappings, and the current gap lists. Execution metadata was normalized so the committed notebook contains no timestamps, local checkout paths, credentials, or transient errors.

haoranpb (dismissed):
Have you read through the results in the notebook, just wondering if that coverage is by design, looks a bit surprising tbh.

> 9 articles covered across 17/130 annotated entries; 177 of 186 articles have zero coverage

Otherwise looks good

haoranpb (approved):

PR #795, opened 2026-08-17
Decouple code-review from the BC PR Review runner

gggdttt:
## Summary

Follow up on [Haoran's review of #774](https://github.com/microsoft/BC-Bench/pull/774#pullrequestreview-4949112712) by separating the `code-review` evaluation contract from the system under test.

- restore `--category code-review` for the generic GitHub Copilot CLI and Claude Code runners, backed by a strict shared `review.json` prompt contract
- expose BC PR Review through dedicated `bcbench run/evaluate pr-review` commands and move its integration to `bcbench.agent.pr_review`
- keep BC PR Review production fidelity in a dedicated workflow while routing the Copilot and Claude workflows through their normal harnesses
- preserve the shared dataset, scorer, result, and leaderboard schemas; existing `agent_name` and model metadata distinguish the runners, so no result migration is needed

The just-released `code-review` runner command remains as a hidden deprecated alias to avoid an immediate CLI break without presenting the category as a runner in public help.

## Validation

- `uv run ruff format <changed Python files>`
- `uv run ruff check <changed Python files>`
- `uv run ty check . --ignore=unresolved-import --exclude "notebooks/"`
- `uv run pytest -q` (738 passed, 2 skipped)


.github/workflows/pr-review-evaluation.yml
```diff
@@ -0,0 +1,172 @@
...
+          node-version: 24
+
+      - name: Install evaluation CLIs
+        uses: $/.github/actions/install-eval-clis
+
+      - name: Checkout BC-ALAgents review engine
+        uses: actions/checkout@v5
+        with:
+          repository: microsoft/BC-ALAgents
+          ref: main
```
haoranpb: [critical] by pointing to `main`, we are not pointing to a specific commit.

So by triggering two runs separatly, we might be getting 2 different versions of `BC-ALAgents`, this is hurting reproducibility.

We should use a commit sha, and start to document that in the releases.

docs/code-review.md
```diff
@@ -27,6 +27,18 @@ Unlike the pass/fail categories, code review is scored with **Precision / Recall
...
 
+## Category and runners
+
+`code-review` is the evaluation contract: it owns the dataset, structured `review.json` output, scorer, result schema, and leaderboard schema. A runner is the system under test. The same entries can be evaluated through the generic GitHub Copilot CLI and Claude Code runners, allowing direct cross-system comparisons under one scorer.
+
+BC PR Review is a separate production-fidelity runner, analogous to BCal for `nl2al`. It runs the BC-ALAgents review engine with BCQuality while remaining fixed to the `code-review` category:
```
haoranpb: [documentation] you probably don't want to use `nl2al` for elaboration, that category does not have a public leaderboard and documentation.

Consider just call the BC PR Review "agent harness", which is probably the correct naming.

PR #800, opened 2026-08-17
Add dedicated BC PR Review evaluation runner

gggdttt:
## Summary

Reapply the decoupled code-review workflow on top of `revert-774-bench/engine-review-runner` so the complete feature can be reviewed as one focused diff.

- Keep `code-review` available to the generic Copilot and Claude runners.
- Add a dedicated `pr-review` runner backed by the production BC PR Review engine and BCQuality.
- Run the engine's complete local, non-posting pipeline and consume its normalized `al-code-review-findings.json` artifact.
- Reject failed, unknown, and malformed engine outcomes instead of scoring them as clean reviews.
- Supply machine-local engine and BCQuality checkout paths through CLI options while retaining BCQuality repo/ref as experiment settings for remote branches and forks.
- Pin BC-ALAgents to `f2ac8704bf8d39000f8002bcf2d287f1f5b5e9ba`, default to `gpt-5.6-luna`, align the minimum severity with the engine's `Medium` default, and run the engine with the dataset entry's target repository context.

## Dependency

CI authentication for local engine reviews depends on [microsoft/BC-ALAgents#57](https://github.com/microsoft/BC-ALAgents/pull/57), whose exact commit is pinned by this PR.

## Validation

- `uv run pytest -q`: 759 passed, 2 skipped
- PowerShell AST parse check for `Prepare-BCQualityRoot.ps1`

## Context

The target branch reverts #774. Using it as the base makes this PR show the full feature rather than only the follow-up decoupling changes from #795.


.github/workflows/pr-review-evaluation.yml
```diff
@@ -0,0 +1,172 @@
...
+          node-version: 24
+
+      - name: Install evaluation CLIs
+        uses: $/.github/actions/install-eval-clis
+
+      - name: Checkout BC-ALAgents review engine
+        uses: actions/checkout@v5
+        with:
+          repository: microsoft/BC-ALAgents
+          ref: main
```
haoranpb: [critical] we are pointing to `main` instead of a specific commit.

This means, if we trigger two runs separatly, we might be getting 2 different versions of `BC-ALAgents` ==> this is hurting reproducibility.

We should use a commit sha, and start to document that in the releases.

docs/code-review.md
```diff
@@ -27,6 +27,18 @@ Unlike the pass/fail categories, code review is scored with **Precision / Recall
...
 
+## Category and runners
+
+`code-review` is the evaluation contract: it owns the dataset, structured `review.json` output, scorer, result schema, and leaderboard schema. A runner is the system under test. The same entries can be evaluated through the generic GitHub Copilot CLI and Claude Code runners, allowing direct cross-system comparisons under one scorer.
+
+BC PR Review is a separate production-fidelity runner, analogous to BCal for `nl2al`. It runs the BC-ALAgents review engine with BCQuality while remaining fixed to the `code-review` category:
```
haoranpb: [documentation] The BCal/ nl2al analogy relies on internal context that public readers may not have. Could we describe BC PR Review directly as an agent harness instead?

src/bcbench/agent/pr_review/agent.py
```diff
@@ -0,0 +1,269 @@
+"""Run the BC-ALAgents review engine (generate half) as a BC-Bench agent.
```
haoranpb: [documentation] consider just call BC PR Review an "agent harness"

src/bcbench/agent/pr_review/agent.py
```diff
@@ -0,0 +1,269 @@
...
+
+_AGENT_OUTPUT_FILE = "agent-output.txt"
+_REVIEW_OUTPUT_FILE = "review.json"
+_PREPARE_BCQUALITY_SCRIPT = Path(__file__).parent / "scripts" / "Prepare-BCQualityRoot.ps1"
+
+
+def _load_pr_review_settings() -> dict[str, Any]:
+    config_file = _config.paths.agent_share_dir / "config.yaml"
+    data = yaml.safe_load(config_file.read_text()) or {}
+    return data.get("pr_review") or {}
```
haoranpb: [recommendation] drop the safe guard here (i.e. `or {}`), for example, if `data.get("pr_review")` fails, the pipeline should fail.

It is a general assumption we make across categories that `config.yaml` exists, it should fail loudly if not.

---

With that assumption, you can probably simplify quite some code.

src/bcbench/agent/pr_review/agent.py
```diff
@@ -0,0 +1,269 @@
...
+def _resolve_gh_token() -> str:
+    token = os.environ.get("GH_TOKEN") or os.environ.get("GITHUB_TOKEN")
+    if token:
+        return token
+    gh = shutil.which("gh")
+    if gh:
+        result = subprocess.run([gh, "auth", "token"], capture_output=True, text=True, check=False)
+        if result.returncode == 0 and result.stdout.strip():
+            return result.stdout.strip()
+    raise AgentError("No GitHub token available for Copilot CLI auth. Set GH_TOKEN or run `gh auth login`.")
```
haoranpb: [recommendation] make the assumption authentication is not required here: 1. in github workflow, it will be present by default; 2. in our local development, we are logged in `gh`

This way, you avoid handling authentication tokens ourselves, less likely to have security issues.

src/bcbench/agent/pr_review/agent.py
```diff
@@ -0,0 +1,269 @@
...
+
+
+def _git(args: list[str], cwd: Path) -> subprocess.CompletedProcess[str]:
+    return subprocess.run(["git", *args], cwd=str(cwd), capture_output=True, text=True, check=True)
+
+
+def _commit_patch_as_head(repo_path: Path) -> None:
```
haoranpb, line 62: [recommendation] See `commit_changes` from `git_operations.py`, it might not be a drop in replacement, but can be easily updated.

Same goes for the rest of git operations in this file, if we could re-use or expand our utility functions there.

src/bcbench/agent/pr_review/agent.py
```diff
@@ -0,0 +1,269 @@
...
+        "COPILOT_REVIEW_AGENT_VERSION": agent_version,
+        "AGENT_MINIMUM_SEVERITY": severity,
+        "GH_TOKEN": gh_token,
+    }
+
+    plugins = [f"bc-review-engine@{_resolve_engine_revision(engine_root)}"]
+    if bcquality_sha:
+        plugins.append(f"BCQuality@{bcquality_sha}")
+    config = ExperimentConfiguration(
+        custom_agent="bc-review-engine",
```
haoranpb: [critical] Here we are mixing the terminology for [Custom Agent](https://docs.github.com/en/copilot/how-tos/copilot-on-github/customize-copilot/customize-cloud-agent/create-custom-agents), in the context of `ExperimentConfiguration`, "Custom Agent" refers to the custom agent for coding agents (e.g. passed via `--agent` flag to GitHub Copilot CLI).

We are using an entire different agent harness here, not custom agent.

---

Same for plugins.

haoranpb: And plugins?

src/bcbench/agent/shared/config.yaml
```diff
@@ -50,6 +50,26 @@ prompt:
     {{task}}
     {% endif %}
 
+  code-review-template: |
```
haoranpb: [nit] is this a mistake or deliberate? if mistake, the diff should be clearer if we move it back

src/bcbench/commands/evaluate.py
```diff
@@ -150,6 +202,47 @@ def evaluate_claude_code(
...
+        bcquality_ref=bcquality_ref,
+        bcquality_repo=bcquality_repo,
+        bcquality_local_path=bcquality_local_path,
+        min_severity=min_severity,
+    )
+
+
+# Compatibility for the command name introduced in v0.8.0. Keep it out of public help
+# because "code-review" names the category, not this specific runner.
+evaluate_app.command("code-review", hidden=True, deprecated=True)(evaluate_pr_review)
```
haoranpb: We intentionally don’t preserve backward compatibility; please remove this deprecated alias and its tests.

---

same for the run command

src/bcbench/commands/evaluate.py
```diff
@@ -150,6 +202,47 @@ def evaluate_claude_code(
     logger.info(f"Results saved to: {run_dir}")
 
 
+@evaluate_app.command("pr-review")
+def evaluate_pr_review(
+    entry_id: Annotated[str, typer.Argument(help="Entry ID to run")],
+    model: CopilotModel = "claude-sonnet-5",
```
haoranpb: Let's default to cheaper models

```suggestion
    model: CopilotModel = "gpt-5.6-luna",
```

---

Same for the run command

src/bcbench/commands/evaluate.py
```diff
@@ -40,6 +40,58 @@ def _prepare_run_dir(output_dir: Path, run_id: str) -> Path:
     return run_dir
 
 
+def _run_pr_review_evaluation(
```
haoranpb: [nit] no need to create a function here, probably easier to move it into where it's called

src/bcbench/commands/run.py
```diff
@@ -26,6 +25,39 @@
 run_app = typer.Typer(help="Run agents on single dataset entry")
 
 
+def _run_pr_review(
```
haoranpb: [nit] no need to create a function here, probably easier to move it into where it's called

---

typer provide a way to directly test commands, see `test_cli_commands.py`, the wrapper is thin here

src/bcbench/commands/evaluate.py
haoranpb: same for the run command

src/bcbench/cli_options.py
```diff
@@ -23,6 +23,7 @@
 
 EvaluationCategoryOption = Annotated[EvaluationCategory, typer.Option(help="Category of evaluation to perform")]
```
haoranpb, line 50: [nit] newline

src/bcbench/types.py
```diff
@@ -193,7 +194,7 @@ def expected_metrics(self) -> frozenset[str]:
                     completion_tokens=None,
                     tool_usage=None,
                 )
-            case AgentHarness.BCAL:
+            case AgentHarness.BCAL | AgentHarness.PR_REVIEW:
```
haoranpb, line 197: [design] since we use `GitHub Copilot CLI` behind the scene, we should be able to collect metrics like token consumption and cost.

You will have to parse and re-emit those metrics from Copilot CLI.

---

This is important because having the ability to compare cost related metrics is important going forward.

haoranpb (commented):
Try to test by running this locally, since it's a brand new path

src/bcbench/agent/shared/config.yaml
gggdttt: Moved the `code-review-template` back to its original location so the diff no longer includes the accidental prompt relocation.

src/bcbench/types.py
gggdttt: Added parsing and export for prompt/completion tokens plus both Copilot cost formats (`Premium requests` and `AI Credits`). A pinned-engine local run successfully parsed 5.9M prompt tokens, 51K completion tokens, and 50.6 AI Credits.

src/bcbench/agent/pr_review/agent.py
gggdttt: Removed the `custom_agent` and `plugins` misuse. The resolved engine and BCQuality inputs now live in a separate `ExecutionProvenance`, which does not turn a baseline into an experiment; only an explicit BCQuality override populates the experiment's `knowledge_base` field.

src/bcbench/commands/evaluate.py
gggdttt: Changed both PR Review commands and the workflow default to `gpt-5.6-luna`.

src/bcbench/agent/pr_review/agent.py
gggdttt: Removed BC-Bench's token discovery and forwarding logic. Local runs now rely on the authenticated Copilot CLI. CI supplies only `COPILOT_GITHUB_TOKEN`; the pinned BC-ALAgents fix forwards it to the child process only for local reviews running under CI.

src/bcbench/agent/pr_review/agent.py
gggdttt: Reused and expanded the shared Git operations. The runner now uses `init_repo`, `has_changes`, and `commit_changes`, while `get_repository_revision` provides deterministic dirty-checkout provenance.

src/bcbench/commands/run.py
gggdttt: Inlined the run flow into the Typer command and replaced the wrapper-level test with a direct `CliRunner` command test.

src/bcbench/commands/evaluate.py
gggdttt: Inlined the evaluation flow into the Typer command and updated the test to invoke the command through `CliRunner`.

docs/code-review.md
gggdttt: Rephrased this to describe BC PR Review directly as a separate agent harness and explain how it shares the `code-review` contract with the generic Copilot and Claude runners.

.github/workflows/pr-review-evaluation.yml
gggdttt: Pinned the workflow to `f2ac8704bf8d39000f8002bcf2d287f1f5b5e9ba`. I also documented the engine-bump/release-note process, bumped BC-Bench to `0.9.0`, and record the exact engine plus filtered BCQuality content in result provenance.

src/bcbench/commands/evaluate.py
gggdttt: Removed the deprecated `code-review` aliases from both `run` and `evaluate`, along with their compatibility tests.

src/bcbench/cli_options.py
gggdttt: Added the missing newline.

src/bcbench/agent/pr_review/agent.py
gggdttt: Removed the fallback dictionaries and optional-key lookups. The runner now indexes the required `pr_review`, `min_severity`, and `bcquality` settings directly, so malformed configuration fails immediately.

src/bcbench/agent/copilot/metrics.py
```diff
@@ -34,7 +34,12 @@ def parse_metrics(output_lines: Sequence[str], session_log_path: Path | None = N
         output_lines: Lines from Copilot CLI stderr output
         session_log_path: Optional path to session log file for tool usage parsing
 
-    Expected output format (v1.0.57):
+    Expected output format (v1.0.81):
```
haoranpb: Forgot you have to handled plain text parsing in my previous comment.. Now reading the code, I think I can do a bigger change to improve this area (leverage structural json over plain text), will handle that in a follow up PR, leave that to me.

You will have to remove all the metrics related changes in other files too.

src/bcbench/commands/evaluate.py
```diff
@@ -150,6 +150,65 @@ def evaluate_claude_code(
     logger.info(f"Results saved to: {run_dir}")
 
 
+@evaluate_app.command("pr-review")
+def evaluate_pr_review(
```
haoranpb, line 156: [Not in Scope] Future only, do not include in this PR, consider support experimentation with al lsp.

Could be some interesting findings to see what if Review Agent has access to al lsp

src/bcbench/agent/shared/config.yaml
```diff
@@ -175,3 +179,23 @@ mcp:
...
+#                  repo/ref). Set local_path to iterate on a local BCQuality checkout without
+#                  pushing (it is copied and filtered; the original is never modified).
+#                  CLI flags (--bcquality-repo/-ref/-local-path) override these.
+pr_review:
+  path: null
+  min_severity: Low
+  bcquality:
+    repo: null
+    ref: null
+    local_path: null
```
haoranpb: [design] the configuration file here is designed to host variables we could experiment with.

For `path`, I think it might be more suited to be a CLI flag instead? e.g. see how [ContainerName](https://github.com/microsoft/BC-Bench/blob/8ae07a7a7c2e0f023a416acefb1a21f199546948/src/bcbench/cli_options.py#L18) is implemented.

Similiar for `local_path`.

Also wondering if `repo` here is needed, I guess `local_path` will have similiar effect, unless it's designed for forks or?

src/bcbench/operations/git_operations.py
```diff
@@ -79,11 +81,62 @@ def fetch_commit_if_missing(repo_path: Path, commit: str) -> None:
...
+    subprocess.run(["git", "init", "-q"], cwd=repo_path, stdout=subprocess.DEVNULL, stderr=subprocess.PIPE, check=True)
+
+
+def has_changes(repo_path: Path) -> bool:
+    result = subprocess.run(["git", "status", "--porcelain"], cwd=repo_path, capture_output=True, encoding="utf-8", text=True, check=True)
+    return bool(result.stdout.strip())
+
+
+def get_repository_revision(repo_path: Path) -> str:
+    """Return HEAD plus a deterministic fingerprint when the worktree is dirty."""
```
haoranpb: This is only relavent when testing things locally.
> when the worktree is dirty.

Even if we record the "revision", there is no way to recover it, because it is local.

So not sure if this is needed, consider dropping

src/bcbench/types.py
```diff
@@ -107,13 +120,26 @@ class ExperimentConfiguration(BaseModel):
     # Plugins loaded for this experiment: "<name>@<revision>" (github) or "<name>@local"
     plugins: list[str] | None = None
 
+    # Knowledge base override used by an agent harness, including its resolved revision
+    knowledge_base: str | None = None
+
+    # Resolved execution inputs. Provenance identifies a baseline but does not make it an experiment.
+    provenance: ExecutionProvenance | None = None
```
haoranpb: [design] by introducing `knowledge_base` and `provenance` here, `ExperimentConfiguration` stop being agent-harness agnostic.

In the new workflow file, where we use `ref` to track the `microsoft/BC-ALAgents`'s version. Can we do similiar things like GitHub Copilot CLI and Claude Code? Instead of recording their version in the `ExperimentConfiguration`, we track them in the release note instead.

src/bcbench/agent/shared/config.yaml
haoranpb: For `min_serverity`, do you want to align here? https://github.com/microsoft/BC-ALAgents/blob/f2ac8704bf8d39000f8002bcf2d287f1f5b5e9ba/agents/ALReviewAgent/scripts/Invoke-CopilotPRReview.ps1#L113-L114

haoranpb (changes requested):
Copilot found a few more things that I didn't verify, but probably worth checking.

`env = {**os.environ, ...}`  at  `agent.py:250-251` will forward all the envrionment variables from the workflow, including `GITHUB_REPOSITORY=microsoft/BC-Bench`.

And if you read it downstream in the engineer, we will have to override it

src/bcbench/types.py
gggdttt: Removed `knowledge_base` and `provenance` from `ExperimentConfiguration`, along with PR Review execution provenance in result serialization, exports, and summaries. The runner now returns an empty agent-harness-agnostic configuration; the workflow pin and release notes remain the version-tracking mechanism.

src/bcbench/commands/evaluate.py
gggdttt: Agreed. AL LSP experimentation remains out of scope and is not included in this PR.

src/bcbench/agent/shared/config.yaml
gggdttt: Moved the engine checkout to `--engine-path` (with `BC_PR_REVIEW_ROOT` env support) and the local BCQuality checkout to `--bcquality-local-path`, and removed both machine-local paths from shared config. I retained BCQuality repo/ref for remote branch and fork experiments, and aligned `min_severity` to the engine's `Medium` default.

src/bcbench/operations/git_operations.py
gggdttt: Removed `get_repository_revision`, the dirty-worktree fingerprint, its exports/tests, and the runtime provenance that consumed it.

src/bcbench/agent/copilot/metrics.py
gggdttt: Removed all PR Review metrics additions, including transcript parsing, Premium Requests/AI Credits fields, exports, aggregation, and related tests. The runner now keeps only the existing wall-clock `execution_time`; structural engine metrics can be added in the follow-up.

gggdttt:
Addressed the remaining follow-up feedback in `6f41e65`:

- Removed the deferred plain-text metrics/cost and generic provenance changes.
- Moved machine-local engine and BCQuality paths to CLI options, retained BCQuality repo/ref for remote branch and fork experiments, and aligned the minimum severity to `Medium`.
- Removed the unrecoverable dirty-worktree fingerprint.
- Kept AL LSP experimentation out of scope.

I also fixed the repository-context issue from the review summary: the runner now explicitly sets `GITHUB_REPOSITORY` to the dataset entry's target repository before invoking BC-ALAgents, rather than inheriting `microsoft/BC-Bench` from the workflow.

haoranpb (dismissed):
Looks great! Let's trigger some (test) runs after this is merged, to make sure things are as expected

haoranpb (approved):

PR #804, opened 2026-08-19
Validate code-review article slugs against BCQuality

gggdttt:
## Problem

`ArticleId` only validates the *shape* of an article reference:

```
^[a-z0-9][a-z0-9-]*/[a-z0-9][a-z0-9-]*$
```

Nothing checks that the article actually exists in BCQuality. A one-character typo passes schema validation and corrupts the coverage report in both directions: the entry is credited to a nonexistent article, while the real article silently remains in `zero_coverage`.

## Change

- Add a dataset test that loads `dataset/codereview.jsonl` and verifies every declared article slug against the BCQuality inventory.
- Check out `microsoft/BCQuality` at immutable commit `841b4e7cab84756b680b66328c2f1ceaae3ec892` in every `lint-and-test` matrix job, after pre-commit and before pytest.
- Set `BCQUALITY_ROOT` to `${{ github.workspace }}/bcquality` for pytest, so the validation runs on both Ubuntu and Windows.
- Keep the test opt-in for ordinary local runs when `BCQUALITY_ROOT` is unset, but fail with an actionable message when `GITHUB_ACTIONS=true` and the root is missing.
- Report unknown slugs together with the entry ids that declare them.

No production code changes; this reuses `collect_declared_articles`, `enumerate_inventory`, and `build_coverage_report` as-is.

## Validation

- All declared article slugs exist at BCQuality commit `841b4e7cab84756b680b66328c2f1ceaae3ec892`.
- Focused test without `BCQUALITY_ROOT`: 12 passed, 1 skipped.
- Focused test with the pinned checkout and `GITHUB_ACTIONS=true`: 13 passed.
- Full suite with the pinned checkout: 741 passed, 2 skipped.
- Repository-wide pre-commit checks, including ruff and ty, pass.


haoranpb (changes requested):
FYI, I believe the test is skipped at the moment in CI

https://github.com/microsoft/BC-Bench/actions/runs/32252706611/job/96067179841?pr=804#step:5:111

gggdttt:
@haoranpb Thanks for catching this. The slug validation now runs in both CI matrix jobs: the workflow checks out BCQuality at immutable commit \841b4e7cab84756b680b66328c2f1ceaae3ec892\ and sets \BCQUALITY_ROOT\ for pytest. Local runs still skip when the root is absent, while GitHub Actions now fails with an actionable error instead of silently skipping. I also confirmed that all currently declared slugs exist at the pinned commit.

tests/test_bcquality_article_coverage.py
```diff
@@ -130,3 +143,25 @@ def test_resolve_prefers_explicit_over_env(self, tmp_path: Path, monkeypatch):
...
+
+    def test_missing_root_fails_in_github_actions(self, monkeypatch):
+        monkeypatch.delenv("BCQUALITY_ROOT", raising=False)
+        monkeypatch.setenv("GITHUB_ACTIONS", "true")
+        with pytest.raises(pytest.fail.Exception, match="BCQUALITY_ROOT must point"):
+            _require_bcquality_root()
+
+
+class TestDatasetArticleSlugs:
+    def test_declared_slugs_exist_in_bcquality(self):
```
haoranpb, line 182:
```suggestion
    @pytest.mark.e2e
    def test_declared_slugs_exist_in_bcquality(self):
```

.github/workflows/CI.yml
```diff
@@ -36,8 +36,17 @@ jobs:
           # Render ruff lint findings as inline PR annotations (CI-only; local runs stay human-readable)
           RUFF_OUTPUT_FORMAT: github
 
+      - name: Checkout BCQuality
```
haoranpb, line 56: Move to a separate job so it runs in parallel, for all e2e tests.

```
uv run pytest -m e2e           # E2E only
```

.github/workflows/CI.yml
```diff
@@ -36,8 +36,17 @@ jobs:
...
+        uses: actions/checkout@v5
+        with:
+          repository: microsoft/BCQuality
+          ref: 841b4e7cab84756b680b66328c2f1ceaae3ec892
```
haoranpb: In CI tests, it is probably acceptable to pull the latest from BCQuality so you don't have to maintain the commit sha here.

haoranpb (changes requested):
Good additionl. However, I think this test is now in a new category of its own: `e2e: tests requiring external repositories or services`, because it requires retrieve data from production repo.

Create a new custom marker in pytest like this, in `pyproject.toml`

```toml
[tool.pytest.ini_options]
markers = [
    "integration",
    "e2e",
]
```

Then you can label your tests with `@pytest.mark.e2e`.

Create a dedicated job in CI to run e2e tests in parallel.

haoranpb (changes requested):
Pushed two small commits:
1. Make `e2e` test to not run by default
2. Run e2e tests in both OS that we support

haoranpb (approved):

ventselartur (approved):

PR #807, opened 2026-08-19
Add PR review performance telemetry and dashboards

dayland:
## What & why

Adds performance telemetry for the dedicated BC PR Review evaluation runner without replacing the engine's raw diagnostic artifact.

## Changes

- Strictly validates the complete BC-ALAgents schema-v1 `_run-metrics.json` artifact, including source, field presence, types, and the canonical no-AL shape.
- Promotes the original BC-Bench performance surface: monotonic execution time, prompt/completion/total tokens, actual model API calls, exact AI credits, and BCQuality knowledge retained/pruned counts.
- Keeps producer-only diagnostics such as cache/reasoning breakdowns, failed/usage-call counts, premium requests, completeness, and malformed-record counts in the raw artifact rather than public result and leaderboard schemas.
- Suppresses all usage-derived public values when malformed records make coverage unknowable. When token usage is incomplete but structurally well formed, suppresses token totals while retaining independently exact API-call and AI-credit values.
- Aggregates the public fields in Code Review summaries and leaderboards and displays them in GitHub, console, and dashboard performance views.
- Uses the shared `AgentMetrics.ai_credits` field introduced on `main` and preserves full precision through summary persistence.
- Keeps the BC-ALAgents producer pinned to `533dd39dfe29218c09e5e31c39c78bb72fa20aa2`.
- Intentionally pins `@github/copilot` to `1.0.79`: `1.0.80` is not published to npm, and `1.0.79` is the version whose structured telemetry contract was validated.

## Validation

- Ruff format and lint
- `ty check . --ignore=unresolved-import --exclude notebooks/`
- Focused PR Review telemetry tests
- 781 tests passed, 2 skipped
- `pre-commit run --all-files`

AB#645219

### DIFF
--- a/.github/actions/install-agent-harnesses/action.yml
+++ b/.github/actions/install-agent-harnesses/action.yml
@@ -1,5 +1,10 @@
 name: Install Agent Harnesses
-description: Install the Claude Code and GitHub Copilot CLIs used by the evaluation workflows
+description: Install the agent harnesses used by the evaluation workflows
+
+outputs:
+  bc-alagents-path:
+    description: Path to the pinned BC-ALAgents checkout
+    value: ${{ steps.bc-alagents-path.outputs.path }}
 
 runs:
   using: composite
@@ -10,4 +15,17 @@ runs:
 
     - name: Install GitHub Copilot CLI
       run: npm install -g @github/copilot@1.0.80
+      shell: pwsh
+
+    - name: Checkout BC-ALAgents
+      uses: actions/checkout@v5
+      with:
+        repository: microsoft/BC-ALAgents
+        ref: 533dd39dfe29218c09e5e31c39c78bb72fa20aa2
+        path: .agent-harnesses/bc-alagents
+        persist-credentials: false
+
+    - name: Export BC-ALAgents path
+      id: bc-alagents-path
+      run: '"path=$(Join-Path $env:GITHUB_WORKSPACE ''.agent-harnesses/bc-alagents'')" >> $env:GITHUB_OUTPUT'
       shell: pwsh
--- a/.github/workflows/pr-review-evaluation.yml
+++ b/.github/workflows/pr-review-evaluation.yml
@@ -109,6 +109,7 @@ jobs:
           node-version: 24
 
       - name: Install evaluation CLIs
+        id: install-harnesses
         uses: $/.github/actions/install-agent-harnesses
 
       - name: Run BC PR Review for entry ${{ matrix.entry }}
@@ -122,6 +123,7 @@ jobs:
 
           uv run bcbench evaluate pr-review "${{ matrix.entry }}" `
             --model "${{ inputs.model }}" `
+            --engine-path "${{ steps.install-harnesses.outputs.bc-alagents-path }}" `
             --repo-path "${{ steps.setup-env.outputs.repo_path }}" `
             --output-dir "${{ env.EVALUATION_RESULTS_DIR }}"
 
--- a/.github/workflows/pr-review-evaluation.yml
+++ b/.github/workflows/pr-review-evaluation.yml
@@ -111,25 +111,17 @@ jobs:
       - name: Install evaluation CLIs
         uses: $/.github/actions/install-agent-harnesses
 
-      - name: Checkout BC-ALAgents review engine
-        uses: actions/checkout@v5
-        with:
-          repository: microsoft/BC-ALAgents
-          ref: 533dd39dfe29218c09e5e31c39c78bb72fa20aa2
-          path: bc-alagents-engine
-          token: ${{ github.token }}
-
       - name: Run BC PR Review for entry ${{ matrix.entry }}
         timeout-minutes: 120
         shell: pwsh
         env:
           COPILOT_GITHUB_TOKEN: ${{ github.token }}
+          GH_TOKEN: ${{ github.token }}
         run: |
           Write-Output "::add-mask::$env:COPILOT_GITHUB_TOKEN"
 
           uv run bcbench evaluate pr-review "${{ matrix.entry }}" `
             --model "${{ inputs.model }}" `
-            --engine-path "${{ github.workspace }}/bc-alagents-engine" `
             --repo-path "${{ steps.setup-env.outputs.repo_path }}" `
             --output-dir "${{ env.EVALUATION_RESULTS_DIR }}"
 
--- a/.github/workflows/pr-review-evaluation.yml
+++ b/.github/workflows/pr-review-evaluation.yml
@@ -115,7 +115,7 @@ jobs:
         uses: actions/checkout@v5
         with:
           repository: microsoft/BC-ALAgents
-          ref: f2ac8704bf8d39000f8002bcf2d287f1f5b5e9ba
+          ref: 533dd39dfe29218c09e5e31c39c78bb72fa20aa2
           path: bc-alagents-engine
           token: ${{ github.token }}
 
--- a/src/bcbench/agent/pr_review/agent.py
+++ b/src/bcbench/agent/pr_review/agent.py
@@ -132,7 +132,9 @@ def _write_review_json(output_dir: Path, repo_path: Path) -> int:
     if outcome == "failed":
         reason = report.get("outcomeReason") or "unknown reason"
         raise AgentError(f"Engine review failed: {reason}")
-    if outcome not in {"completed", "partial", "not-applicable", "no-knowledge"}:
+    if outcome == "not-applicable":
+        raise AgentError("Engine review was not applicable. BC-Bench code-review entries must contain AL changes.")
+    if outcome not in {"completed", "partial", "no-knowledge"}:
         raise AgentError(f"Engine {_FINDINGS_OUTPUT_FILE} has unsupported outcome {outcome!r}.")
     if not isinstance(report.get("findings"), list):
         raise AgentError(f"Engine report in {_FINDINGS_OUTPUT_FILE} has no findings list (got {type(report.get('findings')).__name__}); refusing to score it as a clean review.")
--- a/src/bcbench/agent/pr_review/agent.py
+++ b/src/bcbench/agent/pr_review/agent.py
@@ -15,11 +15,13 @@ import os
 import shutil
 import subprocess
 import time
+from hashlib import sha256
 from pathlib import Path
 from typing import Any
 
 import yaml
 
+from bcbench.agent.pr_review.metrics import build_pr_review_metrics
 from bcbench.agent.pr_review.review_output import engine_report_to_review_comments, load_engine_report
 from bcbench.config import get_config
 from bcbench.dataset import BaseDatasetEntry
@@ -27,7 +29,7 @@ from bcbench.dataset.codereview import CodeReviewEntry
 from bcbench.exceptions import AgentError, AgentTimeoutError
 from bcbench.logger import get_logger
 from bcbench.operations import commit_changes, has_changes, init_repo
-from bcbench.types import AgentMetrics, EvaluationCategory, ExperimentConfiguration
+from bcbench.types import AgentMetrics, EvaluationCategory, ExperimentConfiguration, RepositoryProvenance
 
 logger = get_logger(__name__)
 _config = get_config()
@@ -57,6 +59,80 @@ def _resolve_pwsh() -> str:
     if not pwsh:
         raise AgentError("PowerShell (pwsh) not found in PATH. The BC-ALAgents engine requires PowerShell 7+.")
     return pwsh
+
+
+def _git_output(root: Path, *args: str) -> str | None:
+    try:
+        return (
+            subprocess.check_output(
+                ["git", "-C", str(root), *args],
+                stderr=subprocess.DEVNULL,
+                text=True,
+                encoding="utf-8",
+            ).strip()
+            or None
+        )
+    except (OSError, subprocess.CalledProcessError):
+        return None
+
+
+def _git_bytes(root: Path, *args: str) -> bytes | None:
+    try:
+        return subprocess.check_output(
+            ["git", "-C", str(root), *args],
+            stderr=subprocess.DEVNULL,
+        )
+    except (OSError, subprocess.CalledProcessError):
+        return None
+
+
+def _dirty_content_hash(root: Path) -> str:
+    metadata = _git_bytes(root, "diff", "HEAD", "--raw", "-z", "--abbrev=40", "--no-renames", "--no-ext-diff")
+    tracked = _git_bytes(root, "diff", "HEAD", "--name-only", "-z", "--no-renames", "--no-ext-diff")
+    untracked = _git_bytes(root, "ls-files", "--others", "--exclude-standard", "-z")
+    if metadata is None or tracked is None or untracked is None:
+        raise AgentError(f"Could not fingerprint dirty repository at {root}.")
+
+    fingerprint = sha256(metadata)
+    changed_paths = sorted(set(tracked.rstrip(b"\0").split(b"\0")) | set(untracked.rstrip(b"\0").split(b"\0")))
+    for raw_path in filter(None, changed_paths):
+        relative_path = os.fsdecode(raw_path)
+        absolute_path = root / relative_path
+        if not absolute_path.exists():
+            object_hash = "deleted"
+        elif absolute_path.is_dir():
+            object_hash = _git_output(absolute_path, "rev-parse", "HEAD")
+        else:
+            object_hash = _git_output(root, "hash-object", "--no-filters", "--", relative_path)
+        if object_hash is None:
+            raise AgentError(f"Could not fingerprint changed path {relative_path} in {root}.")
+        fingerprint.update(raw_path)
+        fingerprint.update(b"\0")
+        fingerprint.update(object_hash.encode())
+    return fingerprint.hexdigest()
+
+
+def _repository_provenance(
+    root: Path,
+    repository: str | None = None,
+    ref: str | None = None,
+    *,
+    is_variant: bool | None = None,
+) -> RepositoryProvenance | None:
+    commit_sha = _git_output(root, "rev-parse", "HEAD")
+    resolved_repository = repository or _git_output(root, "config", "--get", "remote.origin.url")
+    resolved_ref = ref or _git_output(root, "branch", "--show-current")
+    dirty = bool(_git_output(root, "status", "--porcelain")) if commit_sha else False
+    if not any((resolved_repository, resolved_ref, commit_sha)):
+        return None
+    return RepositoryProvenance(
+        repository=resolved_repository,
+        ref=resolved_ref,
+        commit_sha=commit_sha,
+        dirty=dirty,
+        content_hash=_dirty_content_hash(root) if dirty else None,
+        is_variant=bool(is_variant or dirty or (is_variant is None and resolved_ref)),
+    )
 
 
 def _commit_patch_as_head(repo_path: Path) -> None:
@@ -199,7 +275,15 @@ def run_pr_review_agent(
         "AGENT_MINIMUM_SEVERITY": severity,
     }
 
-    config = ExperimentConfiguration()
+    config = ExperimentConfiguration(
+        pr_review_engine=_repository_provenance(engine_root),
+        bcquality=_repository_provenance(
+            bcquality_local_path or bcquality_root,
+            bcquality_repo,
+            bcquality_ref,
+            is_variant=bool(bcquality_repo or bcquality_ref or bcquality_local_path),
+        ),
+    )
 
     start = time.monotonic()
     try:
@@ -229,4 +313,4 @@ def run_pr_review_agent(
         logger.exception("Unexpected error running engine review")
         raise
     else:
-        return AgentMetrics(execution_time=time.monotonic() - start), config
+        return build_pr_review_metrics(output_dir, time.monotonic() - start), config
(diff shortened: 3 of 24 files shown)


gggdttt (dismissed):

.github/actions/install-agent-harnesses/action.yml
```diff
@@ -9,5 +9,5 @@ runs:
       shell: pwsh
 
     - name: Install GitHub Copilot CLI
-      run: npm install -g @github/copilot@1.0.80
+      run: npm install -g @github/copilot@1.0.79
```
haoranpb: Changed by mistake?

src/bcbench/types.py
```diff
@@ -76,9 +76,26 @@ class AgentMetrics(BaseModel):
...
 
+    # Structured usage metrics emitted by agent harnesses
+    cached_tokens: int | None = Field(default=None, ge=0, exclude_if=lambda value: value is None)
+    cache_creation_tokens: int | None = Field(default=None, ge=0, exclude_if=lambda value: value is None)
+    reasoning_tokens: int | None = Field(default=None, ge=0, exclude_if=lambda value: value is None)
+    total_tokens: int | None = Field(default=None, ge=0, exclude_if=lambda value: value is None)
+    api_calls: int | None = Field(default=None, ge=0, exclude_if=lambda value: value is None)
+    failed_api_calls: int | None = Field(default=None, ge=0, exclude_if=lambda value: value is None)
+    usage_api_calls: int | None = Field(default=None, ge=0, exclude_if=lambda value: value is None)
+    ai_credits: float | None = Field(default=None, ge=0, exclude_if=lambda value: value is None)
```
haoranpb: Could you merge/rebase `main` and resolve `ai_credits` here? #810 already added this field.

src/bcbench/types.py
```diff
@@ -76,9 +76,26 @@ class AgentMetrics(BaseModel):
...
+    premium_requests: float | None = Field(default=None, ge=0, exclude_if=lambda value: value is None)
```
haoranpb: We do not keep backwards compatbility in BC-Bench, prefer ai_credits over the legacy premium requests, unless we need it for a specific reason?

src/bcbench/types.py
```diff
@@ -76,9 +76,26 @@ class AgentMetrics(BaseModel):
...
+    usage_complete: bool | None = Field(default=None, exclude_if=lambda value: value is None)
+    malformed_records: int | None = Field(default=None, ge=0, exclude_if=lambda value: value is None)
+
     # Tool usage statistics from agent logs
     tool_usage: dict[str, int] | None = None
 
+    # BC PR Review's structural BCQuality filter metrics
+    knowledge_files: int | None = Field(default=None, ge=0, exclude_if=lambda value: value is None)
+    knowledge_pruned: int | None = Field(default=None, ge=0, exclude_if=lambda value: value is None)
```
haoranpb: [design] Reading the code that calculate these two fields, are they not deterministic basing on the commit sha from BCQuality? If so, recording that commit sha is enough to track those metrics?

You can then move the code to conduct analysis in a notebook or something.

src/bcbench/types.py
```diff
@@ -76,9 +76,26 @@ class AgentMetrics(BaseModel):
     prompt_tokens: int | None = None
     completion_tokens: int | None = None
 
+    # Structured usage metrics emitted by agent harnesses
```
haoranpb: All those metrics are calculated by reading a file PR Review Agent left behind, so it is not entirely clear what each of those mean: e.g. `usage_complete` and `malformed_records`.

Add comments to elaborate.

---

Also, some metrics might be able to consolate. e.g. is `turn_count` and `api_calls` the same?

haoranpb (changes requested):
In general, I think this is great and fills a gap. The PR Review Agent currently gives us very little performance data beyond wall time, meaning we can't compare with baseline (using Copilot CLI) for performance metrics.

My main concern is that we may be promoting every available signal into the result and leaderboard schemas without a clear decision we expect each metric to support. For the headline comparison, I think the useful set is quality/F1, wall time, and AI credits, potentially with valid-output rate. The token breakdown, API/failed/usage-call counts, completeness, and malformed-record fields feel more like diagnostics that could remain in the raw artifact or CI summary.

src/bcbench/types.py
gggdttt: Removed `premium_requests` from `AgentMetrics`, persisted results, summaries, leaderboards, dashboards, and documentation. The field remains only in the private strict schema-v1 parser because the pinned producer contract requires all 18 raw fields; BC-Bench no longer promotes it.

.github/actions/install-agent-harnesses/action.yml
gggdttt: This is intentional. `@github/copilot@1.0.80` is not published to npm, so a clean install fails with `E404`/`ETARGET`. Version `1.0.79` is published and is the version used to validate the structured telemetry contract. The PR description now records the reason for the pin.

haoranpb: I can find the `@github/copilot@1.0.80` on npm.

If you are searching for it on your local machine, it will probably fail because network restrictions on our local dev machine

dayland:
Addressed the overall schema-design feedback in c025bee.

I narrowed the public performance surface to wall time, prompt/completion/total tokens, and AI credits, alongside the existing quality metrics and valid-output rate. Tokens remain public because they are a key efficiency metric for comparing models and harnesses independently of current credit pricing.

API-call details, knowledge-filter counts, token subcategories, completeness/malformed-record diagnostics, and producer metadata are still strictly validated against schema v1 but remain only in the raw `_run-metrics.json` artifact. Engine and BCQuality provenance are retained publicly using sanitized repository identifiers and resolved commit SHAs; dirty local sources also include a deterministic content fingerprint.

This leaves PR Review populating the same shared telemetry concepts we intend to standardize across BC-Bench (`execution_time`, prompt/completion/total tokens, and `ai_credits`) without expanding this PR into cross-category dashboard standardization. That broader alignment can follow separately.

.github/actions/install-agent-harnesses/action.yml
dayland: Agreed. Restored `@github/copilot@1.0.80` in f82d901 after confirming the package is published on npm, and removed the obsolete documentation claiming it was unavailable. The earlier install failure was environmental rather than a package-version issue.

src/bcbench/types.py
dayland: Resolved by rebasing onto the changes from #810. PR Review now uses the shared `AgentMetrics.ai_credits` field from `main`; there is no duplicate PR Review-specific credit field.

src/bcbench/types.py
dayland: Agreed and addressed in c025bee. `knowledge_files` and `knowledge_pruned` are no longer promoted into `AgentMetrics`, result summaries, leaderboards, or dashboards. BCQuality repository/commit provenance remains public, while filter and knowledge diagnostics remain available in the raw artifacts for notebook or deeper analysis.

src/bcbench/types.py
dayland: Addressed in c025bee by removing these producer diagnostics from the public BC-Bench schema rather than documenting them as headline metrics. `usage_complete`, `malformed_records`, API-call counts, and related fields remain only in the strictly validated raw `_run-metrics.json` contract. `turn_count` and `api_calls` are not equivalent—turns represent agent conversation turns, while API calls can include nested model calls and retries—but neither distinction needs to be surfaced for PR Review in the public leaderboard.

src/bcbench/agent/pr_review/metrics.py
```diff
@@ -0,0 +1,88 @@
...
+RUN_METRICS_FILE_NAME = "_run-metrics.json"
+_NonNegativeInt = Annotated[int, Field(ge=0)]
+_NonNegativeFloat = Annotated[float, Field(ge=0)]
+
+
+class _RunMetrics(BaseModel):
+    model_config = ConfigDict(extra="forbid", frozen=True, strict=True)
+
+    schema_version: Literal[1]
+    metrics_source: Literal["copilot-cli-otel", "not-applicable"]
```
haoranpb, line 20: [documentation] Can you add comments explaining in which scenario we will get "not-applicable" here?

Suspect it is related to implementation in the PR Review Agent but not entirely clear here.

src/bcbench/agent/pr_review/agent.py
```diff
@@ -59,6 +61,80 @@ def _resolve_pwsh() -> str:
...
+    try:
+        return subprocess.check_output(
+            ["git", "-C", str(root), *args],
+            stderr=subprocess.DEVNULL,
+        )
+    except (OSError, subprocess.CalledProcessError):
+        return None
+
+
+def _dirty_content_hash(root: Path) -> str:
```
haoranpb: this is very similar to https://github.com/microsoft/BC-Bench/pull/800#discussion_r3819584011

Most likely due to unclear documentation from BC-Bench, making a change in #819 .

---

The purpose of recording commit sha is to ensure reproducibility, however, that requires the **commit sha exists on "remote"**.

Dirty content sha only happens on local machine, most likely for smoke tests before pushing, it is not needed for ensure reproducibility for this scenario, the commit sha also can't be recovered.

The related changes can be dropped here.

src/bcbench/results/codereview.py
```diff
@@ -474,5 +509,9 @@ def from_results(cls, results: Sequence[BaseEvaluationResult], run_id: str) -> "
                 "severity_mae": round(severity_mae, 3),
                 "valid_review_output_rate": round(valid_output_rate, 3),
                 "instance_results": {r.instance_id: round(r.f1, 6) for r in code_review_results},
+                "average_prompt_tokens": average_metric("prompt_tokens"),
```
haoranpb: Prefer referring those varibles directly over referencing by name, so commands like Find Reference and Rename would still work.

src/bcbench/results/leaderboard.py
```diff
@@ -184,6 +194,11 @@ def from_runs(cls, runs: Sequence[EvaluationResultSummary]) -> "CodeReviewLeader
                 "macro_f_beta_2": sum(r.macro_f_beta_2 for r in cr_runs) / n,
                 "macro_precision": sum(r.macro_precision for r in cr_runs) / n,
                 "macro_recall": sum(r.macro_recall for r in cr_runs) / n,
+                "valid_review_output_rate": sum(r.valid_review_output_rate for r in cr_runs) / n,
+                "average_prompt_tokens": mean_metric("average_prompt_tokens"),
```
haoranpb: Prefer referring those varibles directly over referencing by name, so commands like Find Reference and Rename would still work.

src/bcbench/results/summary.py
```diff
@@ -123,10 +123,9 @@ def from_json(cls, payload: dict[str, Any]) -> "EvaluationResultSummary":
     def to_dict(self) -> dict[str, Any]:
         data = self.model_dump(mode="json")
         data["average_duration"] = round(data["average_duration"], 1)
-        data["average_prompt_tokens"] = round(data["average_prompt_tokens"], 1)
-        data["average_completion_tokens"] = round(data["average_completion_tokens"], 1)
+        data["average_prompt_tokens"] = round(data["average_prompt_tokens"], 1) if data["average_prompt_tokens"] is not None else None
+        data["average_completion_tokens"] = round(data["average_completion_tokens"], 1) if data["average_completion_tokens"] is not None else None
         data["average_llm_duration"] = round(data["average_llm_duration"], 1) if data["average_llm_duration"] is not None else None
```
haoranpb: Is this line removed by mistake?

```python
        data["average_ai_credits"] = round(data["average_ai_credits"], 2) if data["average_ai_credits"] is not None else None
```

src/bcbench/types.py
```diff
@@ -110,13 +165,39 @@ class ExperimentConfiguration(BaseModel):
     # Plugins loaded for this experiment: "<name>@<revision>" (github) or "<name>@local"
     plugins: list[str] | None = None
 
+    pr_review_engine: RepositoryProvenance | None = None
```
haoranpb: [design] I think the PR Review Agent is special here in the sense that the agent is a "real" agent harness, in the same level as Copilot CLI and Claude Code.

And we track the updates to agent harness using release ntoes, for example, in release [v0.9.0](https://github.com/microsoft/BC-Bench/releases/tag/v0.9.0), we specifically mention:
> GitHub Copilot CLI 1.0.80

I would imagine it will be similar how we track the updates in PR Review Agent, instead of a version, we use commit sha from `BC-ALAgent` repo. And that commit sha for PR Review Agent will also pin the specific commit of BCQualiy behind the scenes.

---

By introducing `pr_review_engine` and `bcquality` directly into `ExperimentConfiguration`, the specific implementation of PR Review Agent Harness is bleeding into a agent-agnostic configuration.

src/bcbench/types.py
```diff
@@ -110,13 +165,39 @@ class ExperimentConfiguration(BaseModel):
...
+    bcquality: RepositoryProvenance | None = None
+
+    @computed_field
+    @property
+    def is_experiment(self) -> bool:
```
haoranpb: this function won't be needed if we do not incluide pr_review_engine and bcquality in ExperimentConfiguration.

Experiment will simply be None if it is not an experiment

haoranpb (changes requested):
Here is the overall flow I imagine, by default, the commit sha should follow the pin we have in production (baseline). And to do an experiment, e.g. checking if a change in PR Review Agent brings any improvement:
1. push your changes in a branch in `BC-ALAgent` repo so the commit sha is publicly avaiable
2. checkout a new branch in `BC-Bench` and update the commit sha to the one we got in step 1
3. queue some runs using the new bracnh from step 2
4. Analysis the results, if it's good, merge the changes into `BC-ALAgent` (new commit sha)
5. [optional] update main commit sha from step 4 if results were good

src/bcbench/agent/pr_review/metrics.py
dayland: Added a field comment in 8db5fae. `not-applicable` is emitted when the BC-ALAgents local preflight finds no `.al` files in the selected diff, skips model execution, and writes the canonical exact-zero usage shape.

src/bcbench/results/codereview.py
dayland: Updated in 8a94dd8. The summary helper now averages supplied typed values, and each call references the `AgentMetrics` field directly so Find References, Rename, and static type checking continue to work.

src/bcbench/results/leaderboard.py
dayland: Updated in 8a94dd8. Leaderboard aggregation now passes direct `CodeReviewResultSummary` field values instead of resolving field names through `getattr`, preserving refactoring and type-checking support.

src/bcbench/agent/pr_review/agent.py
dayland: Agreed and removed in 7b8939c. Dirty working-tree provenance and content fingerprints are gone. Publishable benchmark experiments now use a remotely available BC-ALAgents repo/ref, with local checkouts retained only as an execution input for smoke testing.

src/bcbench/types.py
dayland: Agreed and addressed in 7b8939c. BC-ALAgents is now the sole PR Review harness boundary: BC-Bench configures its repo/ref/local source, while BC-ALAgents owns the BCQuality pin through its own configuration. PR-specific engine and BCQuality fields have been removed from the agent-agnostic `ExperimentConfiguration`, result metadata, and dashboard.

src/bcbench/types.py
dayland: Removed in 7b8939c. With PR Review provenance no longer stored in `ExperimentConfiguration`, the extra `is_experiment` property is unnecessary; baseline behavior is again represented by an empty configuration becoming `None`.

dayland:
Implemented the agreed harness-boundary model in 7b8939c and rebased it onto the latest `main` changes from #819:

- BC-ALAgents is the only PR Review harness source configured by BC-Bench (`repo`, `ref`, or local checkout).
- The reproducible baseline pin now lives under `pr_review.engine` in shared config.
- BC-ALAgents exclusively owns its BCQuality repository/ref through its own pinned configuration; direct BCQuality overrides were removed from BC-Bench.
- Ambient `BCQUALITY_*` variables are stripped so they cannot silently override the pinned dependency.
- PR-specific provenance, dirty content hashes, and `is_experiment` were removed from the agent-agnostic experiment/result/dashboard schemas.
- CI now clones the configured harness through the runner, with `GH_TOKEN` available for authenticated sources.

The supported experiment chain is now immutable and production-shaped: BC-Bench commit -> BC-ALAgents commit -> BCQuality commit. Pipeline-only, BCQuality-only, and combined changes are represented by a BC-ALAgents experiment commit.

src/bcbench/results/bceval_export.py
```diff
@@ -24,7 +24,7 @@ def _experiment_metadata(experiment: ExperimentConfiguration | None, git_ref: st
     is_experiment: bool = experiment is not None and not experiment.is_empty()
     return {
         "EvalRunType": "experiment" if is_experiment else "baseline",
-        "experiment": experiment.model_dump(mode="json") if (is_experiment and experiment) else None,
+        "experiment": experiment.model_dump(mode="json") if is_experiment and experiment else None,
```
haoranpb: readability

```suggestion
        "experiment": experiment.model_dump(mode="json") if (is_experiment and experiment) else None,
```

src/bcbench/agent/pr_review/metrics.py
haoranpb: Is this an applicable scenario during evaluation? I assume we will be focus on scenarios where changes do contain with `.al` files. All tasks in our current dataset touch al files/

---

Mostly concern this might accidentally happen behind the scene and causing aggregated metrics to be wrong, without us noticing.

We should fail loadly if that happens, so failure doesn't hide behind a green job.

.github/workflows/pr-review-evaluation.yml
```diff
@@ -111,25 +111,17 @@ jobs:
...
 
-      - name: Checkout BC-ALAgents review engine
-        uses: actions/checkout@v5
-        with:
-          repository: microsoft/BC-ALAgents
-          ref: f2ac8704bf8d39000f8002bcf2d287f1f5b5e9ba
-          path: bc-alagents-engine
-          token: ${{ github.token }}
-
       - name: Run BC PR Review for entry ${{ matrix.entry }}
```
haoranpb, line 115: [design] move the BC-ALAgent setup into the reusable action, this way all agent harnesses' version (commit sha) will be tracked in the same place.

```
$/.github/actions/install-agent-harnesses
```

This keeps all harness pins in one maintainer surface and leaves the Python runner responsible for execution rather than dependency installation. Local runs can continue supplying  `--engine-path`

---

The workflow will have to output the `--engier-path` to be consumed in the later step like before.

src/bcbench/commands/run.py
```diff
@@ -100,10 +99,9 @@ def run_pr_review(
     model: CopilotModel = "gpt-5.6-luna",
     repo_path: RepoPath = _config.paths.testbed_path,
     output_dir: OutputDir = _config.paths.evaluation_results_path,
-    engine_path: PRReviewEnginePath = None,
-    bcquality_ref: Annotated[str | None, typer.Option(help="Override the BCQuality ref (defaults to the engine's pinned ref)")] = None,
-    bcquality_repo: Annotated[str | None, typer.Option(help="Override the BCQuality repo, e.g. a private fork (defaults to config/engine)")] = None,
-    bcquality_local_path: BCQualityLocalPath = None,
+    engine_repo: Annotated[str | None, typer.Option(help="Override the BC-ALAgents repository (defaults to config)")] = None,
```
haoranpb: The `engine_path` parameter and its downstream logic can be kept with the new design.

If we move the agent harness installation into `$/.github/actions/install-agent-harnesses`, refer to the comment on the workflow file.

src/bcbench/agent/shared/config.yaml
```diff
@@ -197,12 +197,11 @@ mcp:
...
+# CLI flags can override this repo/ref or use a local checkout for harness experiments.
+# Local checkouts are for smoke tests only; comparison runs require remote commits.
 pr_review:
   min_severity: Medium
-  bcquality:
-    repo: null
-    ref: null
+  engine:
+    repo: microsoft/BC-ALAgents
+    ref: 533dd39dfe29218c09e5e31c39c78bb72fa20aa2
```
haoranpb: The ref can be tracked in `$/.github/actions/install-agent-harnesses`, refer to the wortkflow file.

haoranpb (changes requested):
Thanks for the iterations, this is much closer. I have two remaining concerns:  `not-applicable` should fail rather than silently enter aggregates for our current dataset, and `BC-ALAgents` should follow the same installation and pinning pattern as the other agent harnesses

src/bcbench/agent/pr_review/metrics.py
dayland: Agreed and changed in 59f005b. BC-Bench still validates the producer contract, but now explicitly fails if either the engine report or `_run-metrics.json` says `not-applicable`. Since every current code-review entry contains AL changes, that outcome is treated as a broken evaluation setup rather than an empty review with zero usage. Added failure tests at both boundaries.

.github/workflows/pr-review-evaluation.yml
dayland: System.Collections.Hashtable.Body

src/bcbench/commands/run.py
dayland: System.Collections.Hashtable.Body

src/bcbench/agent/shared/config.yaml
dayland: System.Collections.Hashtable.Body

haoranpb (approved):

PR #814, opened 2026-08-24
Add extensibility request advisor evaluation

AleksandricMarko:
## Purpose

Adds a single-shot BC-Bench proxy for the interactive extensibility advisor. Given all requester context upfront, the agent classifies the request, checks blockers and existing extensibility points, and either drafts an issue preview, recommends an existing point, or rejects a blocked request. It runs offline and never submits a GitHub issue.

Includes the advisor skill and custom agent, structured `advisor_result.json` output, 10 scenarios, workflow wiring, and tests.

Fixes AB#613841


CATEGORIES.md
haoranpb, line 1: [nit] no need to keep updating this documentation here since it's similiar to pre-existing extensibility ones

haoranpb (approved):
Looks good as the first step, let's do some runs and analysis

PR #817, opened 2026-08-25
Expand extensibility request datasets

AleksandricMarko:
## Summary

Adds additional dataset inputs for the two extensibility request categories:

- 13 new extensibility request triage entries
- 27 new extensibility request implementation entries with corresponding problem statements


src/bcbench/agent/shared/config.yaml
```diff
@@ -119,7 +119,7 @@ prompt:
 # NOTE: the canonical source file is AGENTS.md; it is automatically renamed
 #       to the agent-specific filename (AgentHarness.instruction_filename) during setup
 instructions:
-  enabled: false
+  enabled: true
```
haoranpb, line 122: Leaving a comment here so we remember to change it back

PR #818, opened 2026-08-25
Align contamination probe with the SWE-Bench Illusion methodology

haoranpb:
## Why

PR #738 shipped a file-path identification probe, but it drifted from [The SWE-Bench Illusion (arXiv:2506.12286)](https://arxiv.org/abs/2506.12286) that it cites:

- asked for **top-k=3** paths instead of one
- scored **basename matches and recall**, neither of which is in the paper
- used an arbitrary **date cutoff** as the contamination signal
- claimed test-generation support while scoring `entry.patch`, not `test_patch`

On the last full run, exact hit@3 was 1/101 — and that one path appeared verbatim in the issue text. So the number we had wasn't measuring much.

## What it looks like now

One-shot probe that follows the paper: the model gets only the repo name and issue text (no repository, no tools), returns **one** path, and scores an exact match against any path in the gold patch. The prompt is the paper's template, adapted from `.py` to `.al`.

Also extracted `invoke_copilot` into `agent/copilot/cli.py`. The agent, the code-review judge, and this probe now share one non-interactive invocation with safe defaults (no tools, no custom instructions), replacing three drifting copies of the argument list.

Net **−930/+442**.

## Notable decisions

- **Bug-fix only.** Test-generation support dropped rather than left half-wired. The code stays generic enough for a future category to plug in.
- **Fails loudly.** A malformed answer raises and fails the job so we notice format drift early. Parsing is lenient about presentation (decorated headers, trailing prose, language tags) and strict about substance. `summarize` now uses plain `needs:` so it's skipped if any entry fails — same pattern as `copilot-evaluation`.
- **No path normalization.** All 128 gold paths across the 101 bug-fix entries are already clean POSIX `.al` paths, so matching is literal.
- **No weekly schedule.** A released model's memorization doesn't change on its own; only a new model or new entries move the number, and both are dispatch-triggered.
- **Docs removed.** Methodology is still experimental — workflow summary and notebooks are enough for now.

## Not done yet

Issue-text leakage (paths/imports in the problem statement) isn't labelled. That belongs on the dataset entry rather than in probe-specific metadata, so it's deliberately left for a follow-up.

## Validation

`pre-commit` (ruff, format, `ty check`) passes; full suite **758 passed, 1 skipped**.


gggdttt (commented):
Reviewed the full three-dot diff and verified the claims empirically rather than by reading alone: I re-ran the probe's exact argument vector through `subprocess` (same code path as `run_filepath_identification`), re-computed the dataset statistics with the repo's own `extract_file_paths_from_patch`, and diffed the deleted implementation against the merge-base.

The refactor itself is good: `invoke_copilot` unifying the agent / judge / probe call sites is a clear win, and `-930` lines is well earned. Two things I want to flag before this lands, plus some methodology notes.

---

## What I verified as correct

- **The prompt is a faithful port.** It matches the `get_file_path_prompt` template in Appendix B.1 of arXiv:2506.12286 verbatim, with only `.py` -> `.al` changed. Nicely done.
- **Dropping gold-side normalization is safe.** Recomputed with `extract_file_paths_from_patch`: 101 entries, 128 gold paths, **0** non-canonical (no backslashes, no `a/` or `./` prefixes, no whitespace). The PR body's claim holds.
- `work_dir` isolation via `tempfile.TemporaryDirectory` is the right intent.
- No dangling references left behind by the doc removal.
- `_format_subprocess_output` uses `getattr(..., None)`, so adding `AgentError` to the `except` tuple is safe.
- The artifact glob matches the `<output_dir>/<run_id>/` layout that `prepare_run_dir` creates.

---

## 1. Two failure-handling regressions that compound

These are both introduced by this PR, and together they mean **one bad entry discards the entire run**.

**(a) `parse_prediction` now raises before anything is persisted.** In `runner.py`, `parse_prediction` (L41) runs *before* `save_identification_result` (L53). When it raises, that entry writes nothing at all — including `raw_output`, which is exactly what you need to diagnose the failure. The previous implementation caught the exception and recorded it in an `error` field, so the result still landed.

**(b) `summarize` lost its `if: always()`.** The old workflow had:

```yaml
if: always() && needs.identify.result != 'skipped'
```

The new one uses a plain `needs: identify`. Since `identify` is a matrix job, a single failing leg fails the job and **skips `summarize` entirely** — so a parse failure on 1 entry silently throws away the aggregate over the other 100 model calls.

I fully agree with the "fail loudly" intent, but right now it fails loudly *and* destroys the batch. Suggestions: persist the result before parsing (or include the truncated `raw_output` in the exception message), and restore `if: always()` on `summarize`.

Related: `parse_prediction` is more brittle than it looks. Verified against the actual regex:

| model output | result |
| --- | --- |
| ` ```\nApp/A.al\n``` ` | `['App/A.al']` |
| inline fence ` ```App/A.al``` ` | **raises** — the regex requires a newline after the fence info |
| a second fenced block after RESPONSE | picks the **wrong** block |
| `` `App/A.al` `` inside the fence | returns the backticked string -> never matches gold -> **silent false negative** |

Anchoring on the `RESPONSE` marker instead of "last fence" would be more robust. The last row is also worth noting for the normalization decision: "all 128 gold paths are clean" only covers the **gold** side; the *prediction* side is where `normalize_path` used to earn its keep, and `test_separator_variant_does_not_match` now locks the non-matching behaviour in. It errs conservative (under-reports contamination), but format noise gets scored as "no memorization".

---

## 2. `--available-tools=` does not actually remove tools (pre-existing, but one line to fix here)

The workflow header states the model gets "no repository contents or tools", and the new test is named `test_invoke_copilot_defaults_to_no_tools_and_no_custom_instructions`. I wanted to confirm that guarantee, so I ran the probe's exact argument vector via `subprocess` with an empty temp cwd, a stripped environment, and a real problem statement:

```
turns: 6 | tool.execution_start: 7
['glob', 'rg', 'rg', 'powershell', 'web_fetch', 'powershell', ...]
```

The model had tools and used them, including `web_fetch`.

Control runs isolate the cause — the flag works fine, it is the **empty value** that is a no-op:

| flag | tools available to the model |
| --- | --- |
| `--available-tools=` (what we pass today) | ~95 (and 7 real calls observed) |
| `--available-tools=view` | exactly 1 (`view`) |
| `--available-tools=none` | 0 (model replies `NONE`) |

**To be fair: this is not introduced by this PR** — the deleted `_run_copilot_context_free` used the same flag with a comment making the same claim, so the assumption has been wrong since the probe was written. But this PR promotes it to the shared default in `invoke_copilot` and adds a test that encodes the incorrect guarantee, so it seems worth fixing here rather than leaving it to be re-discovered.

Why it matters for the metric specifically: `identify` runs `actions/checkout`, so `$GITHUB_WORKSPACE` contains `dataset/bcbench.jsonl` — whose `patch` field *is* the answer — and `web_fetch` can reach the upstream repo directly. The `work_dir` temp dir does not constrain `glob` / `rg` / shell. Today's reading is a floor value so nothing is wrong yet, but the moment a model scores higher we cannot tell memorization from lookup, which inverts the purpose of the probe.

Fix appears to be a one-liner: `--available-tools=none`. I'd suggest asserting the observed behaviour (`tool.execution_start == 0` in a real run) rather than asserting the argument string, since the string assertion is what made this look safe.

---

## 3. Methodology notes (discussion, not blocking)

**The control set is gone.** In the paper the contamination signal is a *differential* — SWE-Bench Verified 76% vs external repositories 53%, up to a 47-point gap — and the conclusion explicitly calls for "temporal controls" and "cross-repository validation". Absolute accuracy is not the signal. The deleted module docstring said the same thing: "the contamination signal is the drop". This PR removes `split_by_cutoff` / `--cutoff` / the control set and reports a single absolute match rate. I agree the date cutoff was a weak proxy and worth replacing, but replacing it with *no* control is a larger departure than the one being fixed, which sits awkwardly with "Align ... with the methodology". The floor value is informative today; it stops being attributable as soon as the number moves.

**The one observed hit is prompt-copying.** I scanned all 101 problem statements: **1** README contains the full gold path verbatim (`microsoft__BCApps-4699`, inside its `## Hints` section), 1 contains the gold filename, and 3 have a `## Hints` section. So the current match rate is entirely leakage-driven. Deferring the labelling is reasonable, but the PR description should note that the current reading is known-contaminated. Worth noting too that the paper feeds the raw `problem_statement`, whereas `get_task()` reads our curated README including `## Hints` — that is an extra departure from the reference setup.

**Multi-path entries.** 19 of 101 entries have more than one gold path, so "matches any" makes them structurally easier to hit. Low priority (the paper is exact-match too), but it makes per-entry comparability loose.

---

## 4. Smaller items

- **`prepare_run_dir` uses `shutil.rmtree`** even though the module it was moved into exists precisely to provide `remove_tree` (`onexc=_force_remove_readonly`, docstring: "usually occur in secondary runs on Windows"). Deleting a previous run directory is exactly that scenario.
- **The judge's stdout fallback changed semantics.** It previously received the full raw stdout; `invoke_copilot` now forces `--output-format=json` and returns the `parse_output` final response. The primary path (reading `result_path`) is unaffected, but the fallback now sees something different and I don't think that's covered.
- **The PR description says "workflow summary and notebooks are enough for now"**, but there is no contamination notebook in `notebooks/` (only bug-fix, test-generation, code-review-coverage, dataset). After deleting the docs page and the `docs/index.md` Diagnostics entry, the methodology lives only in a YAML comment and a module docstring.
- **Import style is inconsistent**: `commands/contamination.py` uses `from bcbench.operations.filesystem_operations import prepare_run_dir`, `commands/evaluate.py` uses `from bcbench.operations import prepare_run_dir`.
- **`find_copilot` moved inside the `try` in `agent.py`**, so a missing CLI now surfaces as "Unexpected error running Copilot CLI" via the broad `except Exception`.
- The result suffix changed from `.file-path-identification.jsonl` to `.filepath-identification.jsonl`. The workflow glob is updated consistently, but previously uploaded artifacts won't load with the new loader — fine if intentional, just noting it.

---

## Question

Has this been run end-to-end via `workflow_dispatch` yet? The description mentions pre-commit and unit tests, and both item 1 and item 2 above are the kind of thing only a real run surfaces. (For what it's worth, runtime is not a concern — a clean run took 6 turns, comfortably inside the 20-minute step timeout.)

src/bcbench/agent/copilot/cli.py
```diff
@@ -0,0 +1,75 @@
...
+    """Run one non-interactive Copilot CLI prompt.
+
+    Returns:
+        A tuple containing parsed agent metrics, when available, and the final assistant response. The response is empty when none is emitted.
+    """
+    copilot_cmd = _find_copilot()
+    if not copilot_cmd:
+        raise AgentError("Copilot CLI not found in PATH. Please ensure it is installed and available.")
+
+    tool_access_arg = "--allow-all-tools" if allow_all_tools else "--available-tools="
```
gggdttt: `--available-tools=` with an empty value is a **no-op** - it does not remove tools.

Verified by running this exact argument vector through `subprocess` (the same path the probe takes), with a stripped environment, an empty temp cwd, and a real problem statement:

    turns: 6 | tool.execution_start: 7
    ['glob', 'rg', 'rg', 'powershell', 'web_fetch', 'powershell', ...]

Control runs isolate the cause to the empty value - the flag itself works fine:

| flag | tools available to the model |
| --- | --- |
| `--available-tools=` (what we pass) | ~95, and 7 real calls observed |
| `--available-tools=view` | exactly 1 (`view`) |
| `--available-tools=none` | 0 (model replies `NONE`) |

To be fair, this is **not introduced by this PR** - the deleted `_run_copilot_context_free` used the same flag with a comment making the same claim. But this promotes it to the shared default for all three call sites, so it seems worth fixing here: `--available-tools=none` appears to be a one-line fix.

src/bcbench/contamination/runner.py
```diff
@@ -1,109 +1,64 @@
...
-    logger.info("Running context-free file-path identification on %s (model=%s, top_k=%d)", entry.instance_id, model, top_k)
-    try:
-        with tempfile.TemporaryDirectory(prefix="bcbench-fpid-") as tmp:
-            raw_output = _run_copilot_context_free(prompt, Path(tmp), model)
-        predicted = parse_prediction(raw_output, top_k)
-    except (subprocess.TimeoutExpired, subprocess.CalledProcessError, OSError, AgentError) as exc:
-        error = f"{type(exc).__name__}: {exc}"
-        logger.exception("File-path identification failed for %s", entry.instance_id)
+    gold_files: list[str] = extract_file_paths_from_patch(entry.patch)
+    predicted_files: list[str] = parse_prediction(raw_output)
```
gggdttt: `parse_prediction` runs *before* `save_identification_result` (L53), so when it raises this entry persists **nothing at all** - including `raw_output`, which is exactly what you need to diagnose the failure. The previous implementation caught the exception into an `error` field, so the record still landed.

Combined with the `summarize` change in the workflow, a parse failure on one entry discards the aggregate over the other 100 model calls. I agree with the fail-loudly intent, but right now it fails loudly *and* destroys the batch.

Suggestion: persist first and validate after, or include a truncated `raw_output` in the exception message.

src/bcbench/contamination/runner.py
```diff
@@ -1,109 +1,64 @@
...
+def run_filepath_identification(entry: BugFixEntry, model: str, result_dir: Path) -> FilePathIdentificationResult:
+    task = entry.get_task()
+    prompt: str = build_identification_prompt(task, repo=entry.repo)
 
-    raw_output = ""
-    error: str | None = None
-    predicted: list[str] = []
+    logger.info("Running context-free filepath identification on %s (model=%s)", entry.instance_id, model)
+    with tempfile.TemporaryDirectory(prefix="bcbench-fpid-") as tmp:
+        metrics, raw_output = invoke_copilot(
```
gggdttt, line 33: `allow_all_tools` isn't passed here, so the probe's "no tools" property depends entirely on the default in `cli.py` L31 - while `agent.py` passes `True` to the same function. Three call sites sharing one default with opposite requirements is fragile; passing `allow_all_tools=False` explicitly would make the probe's core guarantee local and readable.

.github/workflows/contamination.yml
```diff
@@ -1,11 +1,7 @@
...
+name: Contamination
 
-# Context-free bug localization task (SWE-Bench Illusion, arXiv:2506.12286):
-# the model sees ONLY the bug report — no repository, no code — and must name the
-# file(s) that contain the bug. Because it needs no BC container or build, this
-# runs cheaply on ubuntu-latest. Interpret the result RELATIVELY: pass a --cutoff
-# so old/public entries are compared against fresh control entries. The delta,
-# not the absolute rate, is the memorization signal.
+# One-shot bug localization probe inspired by The SWE-Bench Illusion (arXiv:2506.12286).
+# The model receives the repository name and bug report, but no repository contents or tools, and predicts one repository-relative path.
```
gggdttt, line 4: "no repository contents or tools" is not what actually happens. `--available-tools=` does not remove tools (details in my comment on `cli.py` L45) - a clean run of this exact configuration made 7 tool calls, including `web_fetch`.

It matters here specifically because `identify` runs `actions/checkout`, so `$GITHUB_WORKSPACE` contains `dataset/bcbench.jsonl` whose `patch` field *is* the answer, and `web_fetch` can reach the upstream repo directly. The `work_dir` temp dir does not constrain `glob` / `rg` / shell.

Today's reading is a floor value so nothing is wrong yet, but the moment a model scores higher we can't separate memorization from lookup - which inverts the purpose of the probe.

.github/workflows/contamination.yml
```diff
@@ -137,14 +117,4 @@ jobs:
...
-          # Manual runs use the dispatch input; scheduled runs fall back to the
-          # CONTAMINATION_CUTOFF repo variable so the weekly report still shows the delta.
-          $cutoff = "${{ inputs.cutoff || vars.CONTAMINATION_CUTOFF }}"
-          $cutoffArg = if ($cutoff) { "--cutoff", $cutoff } else { @() }
-
-          uv run bcbench contamination summarize `
-            --results-dir "${{ env.IDENTIFICATION_RESULTS_DIR }}" `
-            --model "${{ inputs.model || 'claude-haiku-4.5' }}" `
-            @cutoffArg
+        run: uv run bcbench contamination summarize --results-dir "${{ env.IDENTIFICATION_RESULTS_DIR }}"
```
gggdttt, line 120: This job lost the guard the old workflow had:

    if: always() && needs.identify.result != 'skipped'

`identify` is a matrix job, so a single failing leg fails the whole job and **skips `summarize` entirely**. Together with `parse_prediction` now raising (see `runner.py` L41), one unparseable response throws away the summary for every other entry. Please restore the guard.

src/bcbench/contamination/filepath_identification.py
```diff
@@ -1,249 +1,86 @@
...
-
-
-class IdentificationAggregate(BaseModel):
-    model_config = ConfigDict(frozen=True)
+    Raises:
+        ValueError: If the model did not answer with exactly one fenced path.
+            The probe is new, so unanswerable output should surface loudly
+            rather than quietly count as a miss.
+    """
+    blocks: list[str] = re.findall(r"```[^\n]*\n(.*?)\n?```", raw_text, re.DOTALL)
```
gggdttt: This regex is more brittle than the docstring suggests. Verified against it directly:

| model output | result |
| --- | --- |
| fenced path on its own line | `['App/A.al']` |
| single-line inline fence | **raises** - a newline after the fence info is required |
| a second fenced block after RESPONSE | picks the **wrong** block |
| a backtick-wrapped path inside the fence | returns the backticked string, so it never matches gold - a **silent false negative** |

Anchoring on the `RESPONSE` marker would be more robust than "last fence".

The last row also bears on the normalization decision: "all 128 gold paths are already clean" only covers the **gold** side (which I confirmed - 128 paths, 0 non-canonical). The *prediction* side is where `normalize_path` used to earn its keep, and `test_separator_variant_does_not_match` now locks the non-matching behaviour in. It errs conservative, but format noise gets scored as "no memorization".

src/bcbench/operations/filesystem_operations.py
```diff
@@ -33,3 +33,12 @@ def clear_directory(path: Path) -> None:
         else:
             child.chmod(stat.S_IWRITE)
             child.unlink()
+
+
+def prepare_run_dir(output_dir: Path, run_id: str) -> Path:
+    """Prepare a directory for a run, removing any existing contents."""
+    run_dir = output_dir / run_id
+    if run_dir.exists():
+        shutil.rmtree(run_dir)
```
gggdttt: This module provides two Windows-safe removal helpers - `remove_tree` (L14, `onexc=_force_remove_readonly`, docstring: "usually occur in secondary runs on Windows") and `clear_directory` (L23) - and this new function bypasses both with a bare `shutil.rmtree`.

Deleting a previous run directory is exactly the "secondary run on Windows" scenario `remove_tree` was written for, so this should probably be `remove_tree(run_dir)`.

src/bcbench/evaluate/codereview_judge.py
```diff
@@ -148,32 +148,18 @@ def judge_verdicts(
...
             timeout=_config.timeout.agent_execution,
-            check=True,
+            allow_all_tools=True,
         )
-    except (subprocess.TimeoutExpired, subprocess.CalledProcessError, OSError) as exc:
+    except (subprocess.TimeoutExpired, subprocess.CalledProcessError, OSError, AgentError) as exc:
         raise LLMJudgeError(f"Judge subprocess failed: {exc}{_format_subprocess_output(exc)}") from exc
 
-    return _parse_judge_results(result_path, len(pairs), stdout=completed.stdout or "")
+    return _parse_judge_results(result_path, len(pairs), stdout=stdout)
```
gggdttt: Semantics changed here: `stdout` used to be the full raw CLI output, but `invoke_copilot` forces `--output-format=json` and returns the `parse_output` final response. The primary path (reading `result_path`) is unaffected, but this fallback now receives something different, and I don't think there's a test covering it.

tests/test_copilot_agent.py
```diff
@@ -3,10 +3,53 @@
...
+            "bcbench.agent.copilot.cli.subprocess.run",
+            return_value=subprocess.CompletedProcess(args=[], returncode=0, stdout='{"type":"result"}\n', stderr=""),
+        ) as mock_run,
+    ):
+        invoke_copilot(prompt="do the task", model="test-model", work_dir=tmp_path, timeout=60)
+
+    assert mock_run.call_args.args[0] == [
+        "copilot",
+        "--output-format=json",
+        "--available-tools=",
```
gggdttt: This assertion is part of why the empty-value bug looks safe: it pins the argument *string*, while the test name claims a "no tools" guarantee that doesn't actually hold (see `cli.py` L45).

Suggest asserting observed behaviour instead - e.g. `tool.execution_start == 0` in a real run - or at minimum renaming it so it doesn't assert a property it can't verify.

gggdttt (commented):
Adding line-level pointers for the findings from my previous review, so they're easier to action. No new findings - same content, just located against the diff.

tests/test_filepath_identification.py
```diff
@@ -31,208 +21,156 @@
...
+        problem_dir = create_problem_statement_dir(tmp_path, "Sales invoice posting fails")
+        monkeypatch.setattr(type(entry), "problem_statement_dir", property(lambda self: problem_dir))
+        metrics = AgentMetrics(execution_time=1.0)
+        raw_output = """DISCUSSION
+This bug concerns the Bar table.
+RESPONSE
+```
+App/Foo/Bar.Table.al
+```"""
+        monkeypatch.setattr(runner_mod, "invoke_copilot", lambda **_kwargs: (metrics, raw_output))
```
gggdttt: `lambda **_kwargs` discards the kwargs, so nothing on the probe side asserts how the CLI is invoked. Given that `run_filepath_identification` relies on the shared default in `cli.py` for its "no tools" property, this is the test that would catch a regression - and it currently can't.

Capturing kwargs and asserting `allow_all_tools is False` would close that gap.

src/bcbench/commands/contamination.py
```diff
@@ -95,55 +62,22 @@ def _pct(value: float) -> str:
...
-        pre, post = split_by_cutoff(results, cutoff)
-        pre_agg = aggregate_results(pre, label=f"pre-{cutoff} (suspect)")
-        post_agg = aggregate_results(post, label=f"{cutoff}+ (control)")
-        lines.append(_aggregate_row(pre_agg))
-        lines.append(_aggregate_row(post_agg))
-
-        exact_delta = pre_agg.exact_hit_rate - post_agg.exact_hit_rate
-        basename_delta = pre_agg.basename_hit_rate - post_agg.basename_hit_rate
-        lines += [
+def _build_markdown_report(results: list[FilePathIdentificationResult]) -> str:
```
gggdttt, line 65: This reports a single **absolute** match rate with no control set.

In the paper the contamination signal is a *differential* - SWE-Bench Verified 76% vs external repositories 53%, up to a 47-point gap - and the deleted module docstring said the same thing: "the contamination signal is the drop". Absolute accuracy isn't the signal.

I agree the date cutoff was a weak proxy and worth replacing, but removing `split_by_cutoff` / `--cutoff` without a replacement control is a larger departure than the one being fixed, which sits awkwardly with "Align ... with the methodology". The floor value is informative today; it stops being attributable the moment the number moves.

Worth surfacing in the report too: I scanned all 101 problem statements and exactly **1** README contains the full gold path verbatim (`microsoft__BCApps-4699`, inside its `## Hints` section), 1 contains the gold filename, and 3 have a `## Hints` section. So the single observed hit is prompt-copying, and the current rate is entirely leakage-driven. Also note the paper feeds the raw `problem_statement`, whereas `get_task()` reads our curated README including `## Hints`.

src/bcbench/commands/contamination.py
```diff
@@ -1,92 +1,59 @@
...
 
 from bcbench.cli_options import CopilotModel, EvaluationCategoryOption, OutputDir, RunId
 from bcbench.config import get_config
-from bcbench.contamination.filepath_identification import FilePathIdentificationResult, IdentificationAggregate, aggregate_results, split_by_cutoff
+from bcbench.contamination.filepath_identification import FilePathIdentificationResult
 from bcbench.contamination.runner import load_identification_results, run_filepath_identification
-from bcbench.dataset import RepoGroundedEntry
+from bcbench.dataset import BugFixEntry
 from bcbench.logger import get_logger
+from bcbench.operations.filesystem_operations import prepare_run_dir
```
gggdttt: Nit: inconsistent with `commands/evaluate.py`, which imports the same symbol as `from bcbench.operations import prepare_run_dir`.

.github/workflows/contamination.yml
haoranpb: This is by design, at least at the current stage, I would like the failures to be as loud as possible for now, so we could identify all the issues

src/bcbench/agent/copilot/cli.py
haoranpb: Great catch

src/bcbench/evaluate/codereview_judge.py
haoranpb: Unfortunatly, this is hard to test. Good catch.

I renamed the return variable so the senmatics is explict.

I wonder if the proper solution here is to drop the fallback machinsm, if the expected file doesn't exists, we fail the job. It'd be very hard to parse the fallback responces anyway

src/bcbench/commands/contamination.py
haoranpb: Good challenge, my immediate goal is narrower, doen't 100% align with the paper: run the same context-free task across models and compare their absolute match rates. Adjusting the documentations.

As for the leakage, I plan to address in a follow up PR, to include a meta data field on the dataset, signaling whether the problem statement (with hints) link the gold file path or not.

src/bcbench/contamination/runner.py
haoranpb: Good catch, we will try catch the error and raise with raw_ouput for troubleshoot, at this stage, we prefer to fail loudly so we iterate

haoranpb:
Thanks for the review! @gggdttt , no, haven't run it yet. 

Need to merge the workflow file change into `main` before workflow can be run...

gggdttt (approved):

PR #841, opened 2026-09-02
experiment: ablate the al-appsource-review leaf (DO NOT MERGE)

gggdttt:
**Experiment arm — DO NOT MERGE.** Opened for tracking and discussion only.

## Question

Does the `al-appsource-review` leaf cost more precision than it returns on the code-review set?

## Motivation

Analysis of baseline run [33517690300](https://github.com/microsoft/BC-Bench/actions/runs/33517690300) (BC PR Review, `gpt-5.6-luna`, 145 entries, precision 0.151 / recall 0.402 / F1 0.219) attributes **169 of the 623 generated comments (27%)** to the AppSource leaf, spread across **87 of 145 entries**. The gold set expects **zero** AppSource findings.

This is not corpus dilution. Across the 11 domains the correlation between per-domain corpus size and mean entry F1 is `r = +0.02`. The two extremes run the opposite way:

| leaf | knowledge articles | comments generated | entries touched | gold expects |
| --- | ---: | ---: | ---: | ---: |
| `al-appsource-review` | 4 | **169** | 87 / 145 | 0 |
| `al-ui-review` | 28 | 8 | 8 / 145 | 4 |

What drives output volume is the leaf's applicability gate, not how much it knows.

## Why the AppSource leaf engages everywhere

Two properties of `microsoft/skills/review/al-appsource-review.md` combine:

1. Its first worklist cue fires **when no namespace declaration is present**. The synthetic fixtures are bare `src/*.al` files with no namespace, so the cue is unconditionally true.
2. Its `not-applicable` escape hatch keys on a token list that includes `field`, `key`, `control`, and `action` — tokens present in essentially every AL diff.

The leaf then asserts an affix violation on fixtures that carry no `app.json`, and therefore have no configured `mandatoryAffixes` to compare against.

## Arm definition

Engine: microsoft/BC-ALAgents@`7f00cca`, which changes only `agents/ALReviewAgent/bcquality.config.yaml`:

- `disabled-skills` → `microsoft/skills/review/al-appsource-review.md`
- `knowledge.deny` → `microsoft/knowledge/appsource/**`

Both levers are required. `al-code-review` pins its leaves in frontmatter and "does not discover sub-skills implicitly", so deleting the leaf file alone leaves the super-skill free to improvise the domain from general model knowledge. Denying the corpus removes the grounding as well.

Verified locally against the pinned BCQuality ref `712dee9`: the filter removes exactly 1 skill and 4 knowledge articles; the remaining 239 articles and 15 other review leaves are untouched.

## Control

This branch is cut from `c4b72341` — the exact commit the baseline run evaluated — so the harness pin is the only variable. The code-review dataset and scoring pipeline are byte-identical between that commit and `main`.

## Prediction

Removing all 169 comments without losing a match puts the ceiling at precision **0.207** and F1 **0.273**, against a baseline F1 of 0.219. That is an upper bound: the judge is semantic, so some AppSource-labelled comments may currently match an expected finding in another domain. The measured value is what decides whether this is worth a real fix.

## What this does and does not establish

This is an ablation, not a proposed product change. The gold set contains no AppSource entries, so part of any gain is simply removing an unmeasured domain. A positive result justifies narrowing the leaf's applicability gate in BCQuality — requiring real AppSource context such as `app.json` with `mandatoryAffixes` or an `AppSourceCop.json` — so production keeps the domain while fixtures stop triggering it.

### DIFF
--- a/.github/actions/install-agent-harnesses/action.yml
+++ b/.github/actions/install-agent-harnesses/action.yml
@@ -21,7 +21,7 @@ runs:
       uses: actions/checkout@v5
       with:
         repository: microsoft/BC-ALAgents
-        ref: 533dd39dfe29218c09e5e31c39c78bb72fa20aa2
+        ref: 7f00ccad5e99a8cbbb73306bfc6526f6ea4d21f2
         path: .agent-harnesses/bc-alagents
         persist-credentials: false
 


gggdttt:
## Smoke run — 4 entries, [33625319691](https://github.com/microsoft/BC-Bench/actions/runs/33625319691)

The ablation holds and the residual confound did not materialise.

**No AppSource comment survived.** This was the open risk: `al-code-review` pins its leaves in frontmatter, and the orchestrator prompt still names `al-appsource` in its construct-gated list (`Invoke-CopilotPRReview.ps1`, WORKLIST PINNING), so the super-skill had both an instruction to cover the domain and no leaf file to do it with. It did not improvise the domain from general model knowledge.

### Same four entries, baseline vs arm

| entry | baseline generated | arm generated | baseline F1 | arm F1 |
| --- | ---: | ---: | ---: | ---: |
| `synthetic__performance-005` | 15 (3 AppSource, 7 Breaking Changes) | **3** | 0.333 | **1.000** |
| `synthetic__security-004` | 4 | 5 | 0.000 | 0.000 |
| `synthetic__obsolete-clean-01` (FP guard) | 2 (both AppSource) | 1 | 0.000 | 0.000 |
| `synthetic__style-clean-04` (FP guard) | 3 (2 AppSource) | 2 | 0.000 | 0.000 |
| **subset total** | **24** | **11** | | |

Subset metrics: precision **0.125 → 0.273**, recall **0.750 → 0.750**, F1 **0.214 → 0.400**. Matched comments are unchanged at 3 — the arm removes noise without dropping a single true finding on this sample.

`synthetic__performance-005` is the clearest case: 15 comments of which 3 matched, down to 3 comments of which 3 matched.

### Two caveats

`performance-005` also lost 7 Breaking Changes comments, which is beyond the ablation's direct scope. That is either run-to-run variance or a knock-on effect of reducing the leaf count; a single run cannot separate the two.

Both FP-guard entries still fail. The arm reduces their comment volume but neither goes silent, so removing this leaf does not by itself fix the guard problem — 36 of 38 FP-guard entries failed at baseline, and that is a separate defect.

n = 4 and the model is non-deterministic. Full 145-entry run: [33625958668](https://github.com/microsoft/BC-Bench/actions/runs/33625958668).

gggdttt:
# Superseded: serial leaf dispatch already fixes what this ablation targets

**TL;DR** — This PR ablates the `al-appsource-review` leaf to cut false positives. @dayland's serial-leaf calibration run does the same thing better, without touching BCQuality at all: it removes 88% of the AppSource comment flood and improves precision, F1, FP-guard silence, and cost simultaneously. I'm parking this PR. Full analysis below so the data isn't lost.

## 1. Baseline run-to-run variance (n=2, identical BC-Bench sha `c4b7234`)

| run | precision | recall | F1 | generated | matched |
|---|---|---|---|---|---|
| [33517690300](https://github.com/microsoft/BC-Bench/actions/runs/33517690300) | 0.151 | 0.402 | 0.219 | 623 | 94 |
| [33550367685](https://github.com/microsoft/BC-Bench/actions/runs/33550367685) | 0.160 | 0.444 | 0.236 | 648 | 104 |

**Aggregate metrics are stable (±0.01 F1); per-instance scores are not.** `synthetic__security-014` scored 0.0 and 1.0 across these two identical-config runs; `synthetic__upgrade-006` scored 0.462 and 0.0. Any conclusion drawn from a handful of instances is noise — including the 4-instance smoke result I posted earlier on this PR, which I withdraw.

## 2. The comparison that matters: serial vs parallel leaf dispatch

Run [33517694402](https://github.com/microsoft/BC-Bench/actions/runs/33517694402) (branch `experiment/pr-review-serial-leaves`) is a clean single-variable A/B against that baseline:

- **Engine**: BC-ALAgents `c2cec6d` = baseline `533dd39` with only the `COPILOT_REVIEW_PARALLEL_LEAVES` default flipped `true` -> `false`. 5 changed lines: a docstring and the default. Nothing else.
- **Bench**: the harness pin and its test assertion only. Dataset identical (145 entries, 234 expected comments), same model `gpt-5-6-luna`, same judge `gpt-5.3-codex`, same `benchmark_version` 0.9.0.

| metric | parallel #1 | parallel #2 | **serial** |
|---|---|---|---|
| precision | 0.151 | 0.160 | **0.259** |
| recall | 0.402 | 0.444 | 0.350 |
| F1 | 0.219 | 0.236 | **0.298** |
| macro F1 | 0.212 | 0.219 | **0.327** |
| F<sub>β=0.5</sub> (precision-weighted) | 0.172 | 0.184 | **0.273** |
| F<sub>β=2</sub> (recall-weighted) | 0.301 | 0.328 | 0.327 |
| **FP-guard silence** | **2/38** | **1/38** | **9/38** |
| generated comments | 623 | 648 | **317** |
| avg prompt tokens | 4.25M | 4.26M | **1.64M** |

"FP-guard silence" = entries with `expected_comments == []` that scored F1 1.0 (correct silence).

Recall drops 17%, and that is a real cost. But on the **recall-weighted** F<sub>β=2</sub> serial ties the better parallel run (0.327 vs 0.328), while on precision-weighted F<sub>β=0.5</sub> it is 48-59% higher. **Serial is not worse on any headline metric, at 39% of the token cost.**

## 3. Why this PR is redundant

Comment production by domain:

| | AppSource | share of all comments |
|---|---|---|
| parallel #1 | 169 | 27% (largest domain) |
| parallel #2 | 191 | 29% (largest domain) |
| **serial** | **23** | **7% (5th)** |

Serial dispatch suppresses 88% of the AppSource flood **with no BCQuality change**. This PR deletes the leaf and its knowledge to achieve a subset of the same effect. It treats the symptom.

## 4. Mechanism: the variable is arbitration, not granularity

The obvious reading is "finer-grained agents are worse." I don't think that's what the data shows.

In parallel mode the orchestrator dispatches each leaf as an isolated child agent and instructs it (`Invoke-CopilotPRReview.ps1`, WORKLIST PINNING) that its worklist is the full file list and it **must not narrow it**. Each leaf then reports independently and **nothing arbitrates between them**. A leaf whose applicability gate is weak fires on every file unchecked — which is exactly the AppSource pathology. Its first worklist cue in `al-appsource-review.md` triggers **when no namespace declaration is present**, and the synthetic fixtures are bare `src/*.al` with no `app.json`, so the cue is unconditionally true and there is no `mandatoryAffixes` to check against.

In serial mode one agent walks the domains in shared context, sees its own accumulated findings, and self-arbitrates.

So the operative difference is **isolation without an arbitration stage**, not granularity per se. A per-leaf architecture with a real aggregation/arbitration pass is untested and could plausibly beat both — that's the experiment worth designing, rather than concluding that decomposition is bad.

## 5. Limits

- Serial is **n=1**. The gap (+0.06 to +0.08 F1) is several times the observed parallel spread (0.017), so it is very likely signal, but a repeat would settle it.
- Recall genuinely drops on the raw metric.
- The gold set contains **no AppSource entries**, so part of the AppSource reduction is removing an unmeasured domain rather than removing pure error. This cuts against my ablation, not against the serial result.
- `main` still defaults to `PARALLEL_LEAVES=true`; this finding has not been acted on.

## 6. Suggested next steps

1. Repeat the serial run once (n=2) to pin its variance.
2. Independently of dispatch mode, tighten the applicability gate in BCQuality's `al-appsource-review.md` to require real AppSource context (an `app.json` carrying `mandatoryAffixes`, or an `AppSourceCop.json`). That is the production-safe version of what this PR ablates, and it should compose with either mode.
3. Park this PR — no merge intended. (Its `lint-and-test` is red because the harness-pin change wasn't mirrored into `tests/test_review_workflows.py`; not worth fixing for a parked arm.)

gggdttt:
Closing this experiment branch, which was explicitly marked DO NOT MERGE. It ablated the al-appsource-review leaf to measure that leaf's contribution; the measurement is done and the branch is conflicted against main. Keeping an ablation branch open only invites an accidental merge.

PR #844, opened 2026-09-02
dataset: follow the IsHandled article rename in BCQuality

gggdttt:
> **CI is red by design until BCQuality PR #132 merges.** The `e2e` job runs `test_declared_slugs_exist_in_bcquality`, which clones live BCQuality and asserts every declared slug resolves to a real article. The new slug only exists once #132 lands, so this job fails until then and will go green on a re-run afterwards. `lint-and-test` is green on both runners. See the Merge order section below.
Follows the article rename in BCQuality PR #132.  ## What changes  `microsoft/knowledge/events/initialize-ishandled-to-false-before-publishing.md` is renamed to `microsoft/knowledge/events/reset-ishandled-only-when-the-value-can-carry-over.md` in BCQuality PR #132. The gold answer `synthetic__events-ishandled-reset-boundary-01` cites the old slug in `expected_comments[0].articles`, so once that rename lands the annotation points at an article that no longer exists and the entry shows up under `unknown_articles` in the coverage report.  Two lines change:  - `dataset/codereview.jsonl` - the cited slug on that one entry. - `notebooks/code-review-coverage.ipynb` - the same slug inside the committed article-coverage table output.  ## Why this is a pure rename, not a semantic change  The renamed article still covers what the gold answer asserts, so the annotation stays correct:  - The gold answer's expected comment relies on `ValidateTemporaryLocation` seeding `IsHandled` from `SalesLine.IsTemporary()` before publishing. The article's Best Practice admits exactly that: the value may come from an input parameter, field, global, *or earlier code seeds it*. - The gold answer also states two do-not-flag cases - `ValidateQuantity`'s fresh local, and `ValidateQuantityAndLocation`'s second raise. The article's Anti Pattern keeps both: do not match a single non-looping raise using a fresh local Boolean, and do not match a later raise reached only after a semantically valid `if IsHandled then exit;` proves the value is false.  No expected comment text, line range, severity, or domain is touched.  ## Notebook output  The article table is rendered by `pandas.DataFrame.to_string`, so the Article column is right-aligned to the longest slug in the table. That longest slug is `breaking-changes/relocating-a-field-to-a-tableextension-is-not-a-deletion` at 73 characters. The old slug is 54 characters and the new one is 57, both well under the column width, so the column width is unchanged and only the padding on this one row shifts by three spaces. All 79 data rows were checked to still terminate the Article field at the same column.  The rest of the output is left as committed - this PR is a slug sync, not a refresh of the coverage snapshot.  ## Merge order  **This PR should merge after BCQuality PR #132.** Today the old slug is the valid one; the new slug only exists once #132 lands. Merging this first would point the annotation at an article that does not exist yet.  ## Validation  - `uv run ruff format notebooks/code-review-coverage.ipynb` - unchanged - `uv run ruff check .` - all checks passed - `uv run ty check . --ignore=unresolved-import --exclude "notebooks/"` - one pre-existing diagnostic in an unrelated red-team module, none in the touched files - `uv run pytest -q` - 871 passed, 2 skipped - Loader round-trip: `CodeReviewEntry.load` returns 145 entries and `declared_articles()` on the touched entry resolves to the new slug


haoranpb:
@gggdttt JFYI, the tests are failing

gggdttt:
@Haoranpb fixed now

haoranpb (approved):

PR #845, opened 2026-09-02
Add fix-bug custom agent experiment arm

ventselartur:
## Why

The bug-fix experiment should explicitly select a dedicated workflow without adding a new per-skill selection mechanism. BC-Bench already supports custom-agent experiment arms through `agents.enabled` and `agents.name`, which copy the profile agents and invoke the harness with `--agent=<name>`.

This adds a standalone `fix-bug` custom agent containing the trimmed offline workflow from `microsoft/BCAppsBugFix`: investigate and plan, implement in a self-correcting loop, and validate with AL tools when available. Work-item fetch, test baseline, Miapp propagation, critique, AL review, and pull-request phases remain excluded so the agent fits BC-Bench's unattended evaluation model.

## Summary

- Added the modular `fix-bug` custom agent to the `microsoft-BCApps` and `microsoftInternal-NAV` profiles
- Configured the experiment with `skills.enabled: false`, `agents.enabled: true`, and `agents.name: fix-bug`
- Kept the standard bug-fix prompt unchanged so workflow selection comes from `--agent=fix-bug`
- Removed the out-of-scope `skills.names` selection and skill-name result metadata changes
- Removed premature category documentation pending experimental results
- Added tests for custom-agent configuration, prompt behavior, and copying the agent's supporting files

## Validation

- `uv run pytest -q`
- `uv run ruff check .`


EXPERIMENT.md
```diff
@@ -22,6 +22,7 @@ All configurations live in [`config.yaml`](src/bcbench/agent/shared/config.yaml)
 |---|---|---|
 | `instructions.enabled` | `false` | Copy the **entire** `instructions/<profile>/` folder (instructions + skills + agents) into the target repo before running the agent |
 | `skills.enabled` | `false` | Copy **only** `instructions/<profile>/skills/` |
+| `skills.names` | `[]` | Allow-list of skill folder names to copy; empty copies every skill in the profile. Name them explicitly to keep an arm attributable to a single skill. A profile holding none of them gets no skills, and the result records the skills that were actually copied |
```
haoranpb: The current design is, to experiment the isolate effect of a skill:
1. Create an experiment branch as described in this file
2. Enabled the skills setting in `config.yaml`
3. Replace the content `skills` folder with the skills you would like to experiment (delete the not relavent existing skills)

---

While it could be a nice addition to allow explicitly selecting the skills, it is out of the scope of this PR and should be a separate discussion

src/bcbench/agent/shared/config.yaml
```diff
@@ -26,6 +26,9 @@ prompt:
     {% endif %}
     - Do NOT commit any changes to the repository
     - Focus on W1 localization, do NOT worry about other localizations (e.g. DK, APAC)
+    {% if skills_enabled %}
```
haoranpb: [design] if we want to explicitly ask Copilot to use the skill, do you think it is an idea to swtich to a custom agent instead? So you can invoke the Copilot CLI with `copilot --agent "fix-bug"`.

---

Even without this prompt change, I am confident the agent will pick up this skill, if not, it would be a good signal that we should iterate on the description of the skill so it seems more attractive to the agent when fixing a but in AL codebase

docs/bug-fix.md
```diff
@@ -7,6 +7,8 @@ title: Bug Fixing - BC-Bench
 
 This category follows the [SWE-Bench](https://www.swebench.com/) methodology. The system is tasked with fixing a bug in the Business Central (AL) codebase given an issue description.
 
+An optional `bc-fix-bug` skill arm is available for this category: a trimmed vendored copy of the skill from [`microsoft/BCAppsBugFix`](https://github.com/microsoft/BCAppsBugFix), reduced to the phases that can run offline — investigate and plan, then a self-correcting fix loop. The work-item fetch, test baseline, Miapp propagation, critique, AL review and pull-request phases are removed, so the arm stays consistent with this category's evaluation, which reverts any agent edit to a test project and scores the fix against the dataset's own test patch. Enable it with `skills.enabled: true` and `skills.names: ["bc-fix-bug"]`; see [EXPERIMENT.md](https://github.com/microsoft/BC-Bench/blob/main/EXPERIMENT.md) for how to run it as an experiment arm.
```
haoranpb: [documentation] let's remove this for now, when we are ready, we will probably add a section similiar to `MCP Server Experimental Configurations`

We will give a description and show case some results there

haoranpb (changes requested):
Great that you added the skills!

As discussed yesterday, let's keep this simple so we can quickly iterate.

Small tip, we could trigger runs with Claude Code where MCP is not blocked

haoranpb:
@ventselartur The Claude Code run finished: https://github.com/microsoft/BC-Bench/actions/runs/33734248178

I will be working on getting more information of what happened during the run (i.e. the transcript)

You can take a look at the results in the meantime

PR #843, opened 2026-09-02
Refresh PR Review baseline and model leaderboard

dayland:
Refreshes the published code-review benchmark boundary agreed by the team.\n\n## Dependency chain\n- BCQuality main: 182180913e6c90b62e6c2a65f13371e53c7adb5e\n- BC-ALAgents main: ^[9d9249eaa25be88388b60106b4cbf8cd7aa8a7d (microsoft/BC-ALAgents#60)\n- GitHub Copilot CLI: 1.0.82, the latest stable release with Gemini 3.7 support\n\n## Model curation\n- Replace Gemini 3.6 Flash with Gemini 3.7 Flash\n- Keep the existing provider set; Grok is intentionally excluded\n- Exclude superseded models, internal-only Sol Fast, and redundant lightweight tiers\n\n## Benchmark boundary\n- Bump BC-Bench from 0.9.0 to 0.10.0 because harness and model changes affect comparability\n- Synchronize model choices across typed CLI options and evaluation workflows\n\nFocused tests and all pre-commit checks pass. The refreshed baseline and performance leaderboards include three GPT-5.6 Luna runs and three Gemini 3.7 Flash runs from the immutable benchmark commit.

PR #849, opened 2026-09-03
Allow evaluating an engine branch and only modified entries

gggdttt:
## Why

Today `pr-review-evaluation.yml` can only score the engine commit hard-coded in
`install-agent-harnesses`, over the whole `codereview.jsonl` dataset.

That makes it impossible to answer the question the self-improvement loop keeps
asking: *does this not-yet-merged BCQuality article actually suppress the false
positive we wrote a gold answer for?* The article only takes effect once
`bcquality.config.yaml` points at it, and that file lives in BC-ALAgents - so
you need to evaluate an engine **branch**, not the pinned baseline.

BC-Bench deliberately isolates BCQuality (`_environment_without_bcquality_overrides`
in `agent/pr_review/agent.py` strips every `BCQUALITY_*` variable, and
`Prepare-BCQualityRoot.ps1` reads the ref only from the engine's own config).
That isolation is correct and is left untouched here. Instead this PR makes the
*engine* selectable, so the normal engine -> BCQuality chain does the rest:

1. open the BCQuality PR with the new article,
2. push a BC-ALAgents branch whose `bcquality.config.yaml` `ref` points at that PR head,
3. run this workflow with `engine-ref` set to that branch.

## What

Two optional `workflow_dispatch` inputs, both defaulting to today's behaviour:

- **`engine-ref`** - BC-ALAgents ref to evaluate. Blank keeps the pinned
  baseline commit. Plumbed through a new `bc-alagents-ref` input on the
  `install-agent-harnesses` composite action, which falls back to the same
  pinned SHA when the value is empty (a composite action's `default` does not
  apply to an explicitly-passed empty string, hence the `||` fallback).
  It is also forwarded through `requeue-evaluation` so repeated runs stay on the
  same engine.

- **`modified-only`** - forwarded to `get-entries.yml`, which already supports
  it but was never given the value. Scores only the dataset entries changed
  versus `origin/main`, which is what you want when validating a handful of new
  gold answers instead of burning credits on all 251.

Because a `modified-only` run is a biased subset, it is excluded from anything
that would treat it as a baseline: `skip-leaderboard` is set, `requeue` is
skipped, and it gets its own concurrency group so it does not queue behind a
full run.

## Validation

- `uv run ruff format` / `ruff check` clean
- `uv run pytest -q` - 871 passed, 2 skipped
- `test_review_workflows.py` updated: the harness test now asserts the baseline
  SHA is still pinned *and* that the override input exists.
- YAML of both changed files parses.

No behaviour change when both inputs are left blank, which is every scheduled
and requeued run today.

### DIFF
--- a/.github/workflows/pr-review-evaluation.yml
+++ b/.github/workflows/pr-review-evaluation.yml
@@ -41,9 +41,19 @@ on:
         required: false
         default: ""
         type: string
+      engine-ref:
+        description: "BC-ALAgents ref to evaluate (branch, tag, or SHA). Leave blank to use the pinned baseline engine."
+        required: false
+        default: ""
+        type: string
+      modified-only:
+        description: "Only evaluate dataset entries changed versus origin/main"
+        required: false
+        default: false
+        type: boolean
 
 concurrency:
-  group: pr-review-evaluation-${{ inputs.test-run && 'test' || 'full' }}
+  group: pr-review-evaluation-${{ inputs.test-run && 'test' || inputs.modified-only && 'modified' || 'full' }}
   cancel-in-progress: false
 
 env:
@@ -62,6 +72,7 @@ jobs:
   get-entries:
     uses: $/.github/workflows/get-entries.yml
     with:
+      modified-only: ${{ inputs.modified-only }}
       test-run: ${{ inputs.test-run }}
       category: code-review
 
@@ -111,6 +122,8 @@ jobs:
       - name: Install evaluation CLIs
         id: install-harnesses
         uses: $/.github/actions/install-agent-harnesses
+        with:
+          bc-alagents-ref: ${{ inputs.engine-ref }}
 
       - name: Run BC PR Review for entry ${{ matrix.entry }}
         timeout-minutes: 120
@@ -148,11 +161,12 @@ jobs:
       mock: ${{ inputs.test-run }}
       category: code-review
       git-ref: ${{ inputs.git-ref || github.ref_name }}
+      skip-leaderboard: ${{ inputs.modified-only }}
     secrets: inherit
 
   requeue:
     needs: [summarize-results, pin-commit]
-    if: ${{ !cancelled() && !failure() && !inputs.test-run }}
+    if: ${{ !cancelled() && !failure() && !inputs.test-run && !inputs.modified-only }}
     uses: $/.github/workflows/requeue-evaluation.yml
     permissions:
       contents: write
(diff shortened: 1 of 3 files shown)


.github/workflows/pr-review-evaluation.yml
```diff
@@ -41,9 +41,19 @@ on:
         required: false
         default: ""
         type: string
+      engine-ref:
+        description: "BC-ALAgents ref to evaluate (branch, tag, or SHA). Leave blank to use the pinned baseline engine."
```
haoranpb: [nit] shorter is easier to run on GitHub Action's UI
```suggestion
        description: "BC-ALAgents ref to evaluate. Leave blank to use the pinned baseline engine."
```

.github/workflows/pr-review-evaluation.yml
```diff
@@ -41,9 +41,19 @@ on:
...
+        required: false
+        default: ""
+        type: string
+      modified-only:
+        description: "Only evaluate dataset entries changed versus origin/main"
```
haoranpb: [nit] same description as dataset validation workflow?
```suggestion
        description: "Only verify modified entries"
```

.github/workflows/pr-review-evaluation.yml
```diff
@@ -148,11 +161,12 @@ jobs:
       mock: ${{ inputs.test-run }}
```
haoranpb: do we want the result in kusto? if not, maybe mock is enough

gggdttt:
Closing this in favour of the repository's established practice.

The BC-ALAgents revision under evaluation is pinned in `.github/actions/install-agent-harnesses/action.yml`, and changing it is done by editing that pin in a pull request — see #843 (*Refresh PR Review baseline and model leaderboard*), and #789, which added a dedicated skill for the harness version bump. Evaluating an unmerged BCQuality change is therefore already possible without a runtime override: point an engine branch's `bcquality.config.yaml` at it and move the pin, or run the evaluation locally.

This pull request instead made the pin overridable at dispatch time (`ref: ${{ inputs.bc-alagents-ref || '<sha>' }}`). That works against #851, which is currently exporting `bc-alagents-commit`, `bcquality-repository`, `bcquality-commit` and `bcquality-version` as harness outputs so every evaluation result carries the exact identity of what produced it. A runtime override lets the recorded pin and the engine that actually ran diverge, and this pull request only propagated an `engine-ref` string rather than the full identity set. The two changes also touch the same three files, so they conflict textually as well as in design.

`modified-only` is the one independently useful piece here, and its consumer was microsoft/BC-ALAgentsInternal#25, which is being closed alongside this. If it is wanted for manual evaluation runs it can be reproposed on its own after #851 merges, without touching the pin.

.github/workflows/pr-review-evaluation.yml
```diff
@@ -41,9 +41,19 @@ on:
         required: false
         default: ""
         type: string
+      engine-ref:
```
haoranpb: While this input option simplifies how we experiment with a new vesrion (SHA) of engine, it breaks a assumption we made in BC-Bench:

We assume the runs from `main` branch is the baseline. We can't distinguish the results in Kusto between engine SHA#1 and engine SHA#2, because they use the same ref and version of BC-Bench.

.github/workflows/pr-review-evaluation.yml
```diff
@@ -41,9 +41,19 @@ on:
...
+        description: "BC-ALAgents ref to evaluate (branch, tag, or SHA). Leave blank to use the pinned baseline engine."
+        required: false
+        default: ""
+        type: string
+      modified-only:
```
haoranpb: `pin-commit` should also be skipped when modify only is true

PR #851, opened 2026-09-03
Add advanced code review metrics view

dayland:
Adds a dedicated advanced Code Review metrics view and restores detailed PR Review diagnostics using the typed metrics contracts introduced by #865.

With #860 merged, `benchmark_version` and `agent_version` remain the canonical benchmark and harness identity fields, and `engine-sha` remains the workflow override for unpublished comparison runs. This PR adds the PR Review-specific analysis layer:

- Extends `PRReviewMetrics` with optional knowledge, sub-skill, and BCQuality lineage fields while keeping shared `AgentMetrics` limited to stable core measurements
- Parses and persists selected schema-v1 usage diagnostics, including cached/cache-creation/reasoning tokens, API-call outcomes, malformed records, and telemetry coverage
- Derives BCQuality article retained, pruned, used, and suppressed counts plus executed and skipped sub-skill counts from review artifacts
- Treats missing optional filter/findings diagnostics as unavailable rather than failing an otherwise valid evaluation, while malformed artifacts still fail explicitly
- Aggregates optional diagnostics using only runs that reported each metric, preserving unavailable historical values as null rather than zero
- Records the actual internal Copilot CLI version from `_run-metrics.json` and resolves BCQuality repository/commit/content version from the engine config and prepared checkout directly into PR Review result artifacts
- Keeps the shared summarization workflow generic; no PR Review-specific provenance inputs are added
- Maps the legacy `bc_alagents_commit` field to `agent_version` when loading already-published PR Review rows
- Keeps the default Code Review leaderboards aligned with other BC-Bench categories and shows detailed lineage only in the advanced engineering view

`premium_requests` remains validated as part of the incoming engine schema but is intentionally not persisted or displayed, matching #865's selected diagnostics contract.

The default PR Review pin moves to BC-ALAgents `fdc02d7020632795810057500d62cff2a61513d7`, which includes the merged microsoft/BC-ALAgents#62 change required to persist sub-results for complete article and executed-sub-skill diagnostics.


dayland:
Marking this draft while the advanced data contract is completed. BC-ALAgents currently logs BCQuality filtering/consumption but does not persist article retained, pruned, cited/used, sub-skill execution, or usage-completeness diagnostics into BC-Bench result artifacts, so this page cannot yet show the requested full performance set.

haoranpb (commented):
@dayland I think lots of the changes here is not needed, if we strictly follow the [versioning policy](https://github.com/microsoft/BC-Bench/blob/main/CONTRIBUTING.md#versioning-policy) of BC-Bench.

A version of BC-Bench, e.g. [v0.10.0](https://github.com/microsoft/BC-Bench/releases/tag/v0.10.0), specifically list the version of all agent harness.

So a fixed version of BC-Bench should point to a fixed version of all agent harnesses, which should then points to fixed versions of bcquality etc

Note: the version of BC-Bench is already collected as part of the results

dayland:
@haoranpb That makes sense as the intended versioning contract. Would you be open to a thin compromise where we keep the exact pins in the persisted result identity and aggregation key, but only show them compactly as an **Evaluation Stack** cell on the dashboard?

The release remains the canonical description, while the abbreviated BC-Bench, BC-ALAgents, and BCQuality identities save analysts from reconstructing the transitive pins for each row. This also prevents accidental aggregation if results are produced from different commits or experimental pins under the same in-development benchmark version.

haoranpb:
> Sun Haoran (@haoranpb) That makes sense as the intended versioning contract. Would you be open to a thin compromise where we keep the exact pins in the persisted result identity and aggregation key, but only show them compactly as an **Evaluation Stack** cell on the dashboard?
> 
> The release remains the canonical description, while the abbreviated BC-Bench, BC-ALAgents, and BCQuality identities save analysts from reconstructing the transitive pins for each row. This also prevents accidental aggregation if results are produced from different commits or experimental pins under the same in-development benchmark version.

@dayland Good point, you pointed out a very valid gap in the current design:
> we can't really experiment with agent harness version.

We never had that need before PR Review Agent, after some thought, what do you think about this #860 ? Could it be an alternative to what we are trying to achieve here? Or at least partly

haoranpb (changes requested):
With #860 merged,can we trim this to persisting and displaying the required diagnostics? `benchmark_version` and `agent_version` already identify the setup, so I’m not sure we need five extra inputs on the shared summarization workflow.

I understand the convenience of showing the BCQuality version, but the recorded BC-ALAgents SHA already pins it. Could we resolve `bcquality-commit` in the notebooks for deeper analysis instead of passing it through the evaluation workflows?

---

Or, alternatively, collect and resolve the dependency chain as part of the evaluation result

src/bcbench/agent/pr_review/metrics.py
```diff
@@ -76,10 +100,69 @@ def _load_run_metrics(path: Path) -> _RunMetrics:
         raise AgentError(f"Engine run metrics artifact {path} does not satisfy schema version 1: {exc}") from exc
 
 
-def build_pr_review_metrics(output_dir: Path, execution_time: float) -> AgentMetrics:
+def _load_filter_report(path: Path) -> _FilterReport:
+    if not path.exists():
+        raise AgentError(f"BCQuality filter report not found at {path}.")
+    try:
```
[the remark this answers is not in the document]
dayland: Addressed in ea89f12e. A missing `_filter-report.json` now leaves `knowledge_pruned` unavailable (`null`) without failing the evaluation; malformed reports still raise an explicit error.

src/bcbench/agent/pr_review/metrics.py
```diff
@@ -76,10 +100,69 @@ def _load_run_metrics(path: Path) -> _RunMetrics:
...
+    except ValidationError as exc:
+        raise AgentError(f"BCQuality filter report {path} has an invalid shape: {exc}") from exc
+
+
+def _count_available_knowledge(bcquality_root: Path) -> int:
+    def is_knowledge_file(path: Path) -> bool:
+        parts = path.relative_to(bcquality_root).parts
+        return len(parts) >= 3 and parts[0].lower() in _KNOWLEDGE_LAYERS and parts[1].lower() == "knowledge"
+
+    return sum(1 for path in bcquality_root.rglob("*.md") if path.is_file() and is_knowledge_file(path))
```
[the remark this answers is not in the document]
dayland: Addressed in ea89f12e. The BCQuality knowledge inventory count is cached by resolved checkout path for the process lifetime, with regression coverage.

src/bcbench/agent/pr_review/metrics.py
```diff
@@ -76,10 +100,69 @@ def _load_run_metrics(path: Path) -> _RunMetrics:
...
+        for item in collection:
+            if not isinstance(item, dict):
+                continue
+            references = item.get("references", [])
+            if not isinstance(references, list):
+                continue
+            for reference in references:
+                if isinstance(reference, dict) and isinstance(reference.get("path"), str) and reference["path"]:
+                    cited.add(reference["path"].replace("\\", "/").lower())
+
```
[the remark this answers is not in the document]
dayland: Addressed in ea89f12e. References are normalized and counted only when they resolve to a `<layer>/knowledge/**/*.md` path; skill/non-knowledge references and traversal paths are excluded.

.github/actions/install-agent-harnesses/action.yml
```diff
@@ -46,3 +58,22 @@ runs:
...
+        $version = [regex]::Match($config, '(?m)^\s*version:\s*["'']?(?<value>[^"''\s]+)["'']?\s*$').Groups['value'].Value
+        if (-not $repo -or -not $commit -or -not $version) {
+          throw "Could not resolve BCQuality identity from $configPath"
+        }
+        $repo = $repo -replace '^https://github.com/', '' -replace '\.git$', ''
+
+        "copilot-cli-version=1.0.82" >> $env:GITHUB_OUTPUT
+        "bcquality-repository=$repo" >> $env:GITHUB_OUTPUT
+        "bcquality-commit=$commit" >> $env:GITHUB_OUTPUT
+        "bcquality-version=$version" >> $env:GITHUB_OUTPUT
```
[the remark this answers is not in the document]
dayland: Addressed in ea89f12e by removing the action provenance outputs entirely. The Copilot CLI version now comes from the engine `_run-metrics.json`, and BCQuality identity is resolved from the engine config plus the actual prepared checkout commit.

dayland:
Addressed the latest review feedback in `ea89f12e`. The five PR Review-specific inputs have been removed from the shared summarization workflow. Advanced-only provenance is now derived inside the evaluation result from `_run-metrics.json`, the engine BCQuality config, and the actual prepared checkout commit. The diagnostic correctness/performance comments were also addressed: missing filter reports produce null diagnostics, inventory counting is cached, and article references are normalized and restricted to knowledge paths.

dayland:
Updated #851 against the typed metrics contracts merged in #865.

- PR Review-only diagnostics and BCQuality lineage now live on `PRReviewMetrics`; shared `AgentMetrics` contains only core fields.
- The PR Review parser and timeout path return the concrete registered type.
- Summary extraction tolerates generic historical Code Review metrics that do not expose the optional PR Review fields.
- `premium_requests` was removed from persisted metrics, aggregate schemas, tests, and the advanced dashboard. It remains validated only as part of the incoming schema-v1 engine artifact.
- The PR description was updated to match the final scope.

The branch is current with `main`, and all local and GitHub checks pass.

docs/code-review.md
```diff
@@ -102,11 +103,12 @@ Unavailable AI credits remain `null` in bceval exports; observed zero remains ze
...
-      <th>Ver</th>
+      <th>Version</th>
     </tr>
   </thead>
   <tbody>
-    {% assign performance_results = site.data.code-review.aggregate | sort: "average_duration" %}
+    {% assign production_results = site.data.code-review.aggregate | where: "agent_name", "BC PR Review" %}
+    {% assign performance_results = production_results | sort: "average_duration" %}
     {% for agg in performance_results %}
     <tr>
```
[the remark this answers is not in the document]
dayland: Fixed in 554a55b7. The production performance loop now applies the same non-experiment condition as the baseline table, with a regression assertion that both production sections retain the filter.

src/bcbench/agent/pr_review/metrics.py
```diff
@@ -76,10 +107,118 @@ def _load_run_metrics(path: Path) -> _RunMetrics:
...
+    for index in range(len(parts) - 2):
+        if parts[index].lower() in _KNOWLEDGE_LAYERS and parts[index + 1].lower() == "knowledge":
+            return "/".join(parts[index:]).lower()
+    return None
+
+
+def _load_engine_diagnostics(path: Path) -> _EngineDiagnostics:
+    if not path.exists():
+        return _EngineDiagnostics(knowledge_used=0, knowledge_suppressed=0, sub_skills_executed=None, sub_skills_skipped=0)
+    try:
```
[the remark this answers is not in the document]
dayland: Fixed in 554a55b7. A missing findings artifact now produces null for all findings-derived diagnostics, while a present empty artifact produces measured zeros. Added parser and null-aware aggregation coverage for that distinction.

src/bcbench/results/summary.py
```diff
@@ -59,6 +71,15 @@ class EvaluationResultSummary(BaseModel, ABC):
...
 
     benchmark_version: str
+    copilot_cli_version: str | None = None
+    bcquality_repository: str | None = None
+    bcquality_commit: str | None = None
+    bcquality_version: str | None = None
+
+    @model_validator(mode="before")
+    @classmethod
+    def restore_legacy_agent_version(cls, payload: object) -> object:
```
haoranpb: It is part of the design that we don't do backwards compatability in BC-Bench, do we have a specific use case here?

src/bcbench/results/summary.py
```diff
@@ -102,6 +136,10 @@ def _base_fields(cls, results: Sequence[BaseEvaluationResult], run_id: str) -> d
             "github_run_id": run_id,
             "experiment": experiment,
             "benchmark_version": get_benchmark_version(),
+            "copilot_cli_version": consistent_metric_value("copilot_cli_version"),
```
haoranpb: Prefer typed reference over string, so functionalities like rename and find references can better be leveraged by coding agents (and human)

src/bcbench/results/summary.py
```diff
@@ -59,6 +71,15 @@ class EvaluationResultSummary(BaseModel, ABC):
     experiment: ExperimentConfiguration | None = None
 
     benchmark_version: str
+    copilot_cli_version: str | None = None
+    bcquality_repository: str | None = None
+    bcquality_commit: str | None = None
+    bcquality_version: str | None = None
```
haoranpb: Those properties are code-review category specific, should be moved into `CodeReviewResultSummary` defined in `results/codereview.py`

src/bcbench/results/leaderboard.py
```diff
@@ -35,6 +35,15 @@ class LeaderboardAggregate(BaseModel, ABC):
     average_duration: float
 
     benchmark_version: str
+    copilot_cli_version: str | None = None
```
haoranpb: Property belongs to code review specific subclass

src/bcbench/results/codereview.py
```diff
@@ -509,9 +528,25 @@ def average_metric(values: Sequence[int | float | None]) -> float | None:
                 "severity_mae": round(severity_mae, 3),
                 "valid_review_output_rate": round(valid_output_rate, 3),
                 "instance_results": {r.instance_id: round(r.f1, 6) for r in code_review_results},
-                "average_prompt_tokens": average_metric([result.metrics.prompt_tokens if result.metrics else None for result in code_review_results]),
-                "average_completion_tokens": average_metric([result.metrics.completion_tokens if result.metrics else None for result in code_review_results]),
-                "average_total_tokens": average_metric([result.metrics.total_tokens if result.metrics else None for result in code_review_results]),
-                "average_ai_credits": average_metric([result.metrics.ai_credits if result.metrics else None for result in code_review_results]),
+                "average_prompt_tokens": average_metric(metric_values("prompt_tokens")),
+                "average_completion_tokens": average_metric(metric_values("completion_tokens")),
```
haoranpb: prefer typed refernce over string

src/bcbench/agent/pr_review/metrics.py
```diff
@@ -76,10 +107,118 @@ def _load_run_metrics(path: Path) -> _RunMetrics:
...
+def _load_bcquality_identity(engine_root: Path, bcquality_root: Path) -> tuple[str, str, str] | None:
+    config_path = engine_root / "agents" / "ALReviewAgent" / "bcquality.config.yaml"
+    try:
+        config = yaml.safe_load(config_path.read_text(encoding="utf-8"))
+        bcquality = config["bcquality"]
+        repository = bcquality["repo"]
+        version = bcquality["version"]
+    except (OSError, TypeError, KeyError, yaml.YAMLError) as exc:
+        logger.warning(f"BCQuality provenance unavailable from {config_path}: {exc}")
+        return None
```
haoranpb: As we are still iterating on the code, consider let unexpected things fail loudly, so those behavior doesn't hide behind a green job

---

Same for few other places in this file

src/bcbench/results/summary.py
haoranpb: Maybe because we need to load the existing code review data? If so, let's do a one time data migration instead

src/bcbench/results/codereview.py
```diff
@@ -509,9 +528,25 @@ def average_metric(values: Sequence[int | float | None]) -> float | None:
...
+                "average_malformed_records": average_metric(metric_values("malformed_records")),
+                "average_knowledge_files": average_metric(metric_values("knowledge_files")),
+                "average_knowledge_pruned": average_metric(metric_values("knowledge_pruned")),
+                "average_knowledge_used": average_metric(metric_values("knowledge_used")),
+                "average_knowledge_suppressed": average_metric(metric_values("knowledge_suppressed")),
+                "average_sub_skills_executed": average_metric(metric_values("sub_skills_executed")),
+                "average_sub_skills_skipped": average_metric(metric_values("sub_skills_skipped")),
+                "token_coverage_rate": sum(value is not None for value in metric_values("total_tokens")) / total_results,
+                "credit_coverage_rate": sum(value is not None for value in metric_values("ai_credits")) / total_results,
+                "usage_complete_rate": sum(value is True for value in metric_values("usage_complete")) / total_results,
```
haoranpb: The baseline runs, i.e. without the PR Review Agent, but Copilot CLI directly, won't have those metrics, which is fine.

But this will calculate 0%, which is probably not the deisgn. Maybe we should allow `None` value here.

src/bcbench/results/summary.py
dayland: System.Collections.Hashtable.Body

src/bcbench/results/summary.py
dayland: System.Collections.Hashtable.Body

src/bcbench/results/summary.py
dayland: System.Collections.Hashtable.Body

src/bcbench/results/leaderboard.py
dayland: System.Collections.Hashtable.Body

src/bcbench/results/codereview.py
dayland: System.Collections.Hashtable.Body

src/bcbench/agent/pr_review/metrics.py
dayland: System.Collections.Hashtable.Body

src/bcbench/results/codereview.py
dayland: System.Collections.Hashtable.Body

src/bcbench/agent/pr_review/metrics.py
```diff
@@ -1,17 +1,66 @@
...
+FINDINGS_FILE_NAME = "al-code-review-findings.json"
+_KNOWLEDGE_LAYERS = {"microsoft", "community", "custom"}
 _NonNegativeInt = Annotated[int, Field(ge=0)]
 _NonNegativeFloat = Annotated[float, Field(ge=0)]
 
 
+class _FilterRemoval(BaseModel):
+    model_config = ConfigDict(extra="ignore", frozen=True)
+
+    kind: Literal["knowledge", "skill"]
```
gggdttt: Could we also accept `knowledge-sample` here? The pinned engine emits this when removing `.good.al` / `.bad.al` files from disabled layers. This happens with the default config too, since the community layer is disabled, so the parser will fail even after a successful review.

We should accept these records but keep counting only `knowledge` entries in `knowledge_pruned`.

gggdttt (changes requested):
Could we also accept `knowledge-sample` here? The pinned engine emits this when removing `.good.al` / `.bad.al` files from disabled layers. This happens with the default config too, since the community layer is disabled, so the parser will fail even after a successful review.


We should accept these records but keep counting only `knowledge` entries in `knowledge_pruned`

src/bcbench/results/summary.py
dayland: Addressed in 0d9c723b. Runtime backward compatibility was removed, and the checked-in Code Review dashboard data was migrated once from `bc_alagents_commit` to `agent_version`.

src/bcbench/results/summary.py
dayland: Addressed in 0d9c723b. Shared metric aggregation now uses typed `AgentMetrics` field references instead of string-based attribute lookup.

src/bcbench/results/summary.py
dayland: Addressed in 0d9c723b. PR Review provenance fields now live on `CodeReviewResultSummary` rather than the shared summary model.

src/bcbench/results/leaderboard.py
dayland: Addressed in 0d9c723b. PR Review provenance was moved from the shared leaderboard aggregate to `CodeReviewLeaderboardAggregate`.

src/bcbench/results/codereview.py
dayland: Addressed in 0d9c723b. Advanced diagnostics are aggregated from a narrowed `list[PRReviewMetrics]` using typed field references.

src/bcbench/agent/pr_review/metrics.py
dayland: Addressed in 0d9c723b. Existing malformed findings artifacts and invalid BCQuality identity/configuration now fail loudly; only genuinely absent optional diagnostics remain unavailable.

src/bcbench/results/codereview.py
dayland: Addressed in 0d9c723b. Runs without PR Review usage-completeness telemetry now preserve `usage_complete_rate` as unavailable instead of reporting a misleading 0%.

src/bcbench/agent/pr_review/metrics.py
```diff
@@ -76,10 +125,103 @@ def _load_run_metrics(path: Path) -> _RunMetrics:
...
+    for index in range(len(parts) - 2):
+        if parts[index].lower() in _KNOWLEDGE_LAYERS and parts[index + 1].lower() == "knowledge":
+            return "/".join(parts[index:]).lower()
+    return None
+
+
+def _load_engine_diagnostics(path: Path) -> _EngineDiagnostics:
+    if not path.exists():
+        raise AgentError(f"Engine findings artifact not found at {path}.")
+    try:
```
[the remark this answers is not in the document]
dayland: Addressed in c3226956. A missing `al-code-review-findings.json` now leaves all findings-derived optional diagnostics unavailable, while an existing malformed artifact still fails explicitly.

src/bcbench/results/codereview.py
```diff
@@ -509,9 +554,25 @@ def average_metric(values: Sequence[int | float | None]) -> float | None:
...
+                "average_knowledge_files": average_metric([metric.knowledge_files for metric in pr_review_metrics]),
+                "average_knowledge_pruned": average_metric([metric.knowledge_pruned for metric in pr_review_metrics]),
+                "average_knowledge_used": average_metric([metric.knowledge_used for metric in pr_review_metrics]),
+                "average_knowledge_suppressed": average_metric([metric.knowledge_suppressed for metric in pr_review_metrics]),
+                "average_sub_skills_executed": average_metric([metric.sub_skills_executed for metric in pr_review_metrics]),
+                "average_sub_skills_skipped": average_metric([metric.sub_skills_skipped for metric in pr_review_metrics]),
+                "token_coverage_rate": sum(metric.total_tokens is not None for metric in metrics) / total_results,
+                "credit_coverage_rate": sum(metric.ai_credits is not None for metric in metrics) / total_results,
+                "usage_complete_rate": sum(metric.usage_complete is True for metric in pr_review_metrics) / len(pr_review_metrics) if pr_review_metrics else None,
```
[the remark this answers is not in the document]
dayland: Addressed in c3226956. `usage_complete_rate` now excludes unavailable values from both numerator and denominator and remains unavailable when no task reports completeness.

src/bcbench/agent/pr_review/metrics.py
dayland: Addressed in c3226956. Filter reports now accept `knowledge-sample` removals, while only `knowledge` removals contribute to `knowledge_pruned`.

haoranpb (approved):
Sorry for the late review, got caught up on livesite.

Looks good but please trigger a run with this branch before you merge to verify everything is working as expected

PR #740, opened 2026-07-13
Add data-query category: AL query-generation benchmark

onbuyuka:
## What

Adds a new **`data-query`** evaluation category: a benchmark for **answering Business Central data
questions through the BC MCP server's Data Query tools**. Given a natural-language question, the agent
must retrieve the real data from a live BC environment using the `bc_data_*` MCP tools and report
exactly what they return — it cannot answer from general knowledge.

The agent writes two files:
- **`answer.json`** — the result rows that answer the question (one JSON object per row);
- **`query.al`** — the single AL `query` object it used.

Scoring is **execution-based** (no LLM judge): `build` = a parseable `answer.json` was produced;
`resolved` (`ResolutionRate`) = the agent's rows match the gold answer (baked `gold_rows`, else the
entry's `gold_query` run live). Rows compare by value — numbers normalized scale-insensitively
(`500` == `500.0`), Code/No. strings verbatim (`"001"` != `"1"`); column names/order ignored;
order-insensitive unless the entry is marked `ordered`.

> Complements the AI Test Toolkit evals in the platform repo: those test the **MCP server**
> end-to-end; this benchmarks **models/agents** on their ability to use it.

## The interesting part: keeping the MCP server the *only* route to the data

The category is only meaningful if the agent answers **through the MCP tools**. Making that true was
the bulk of the work — denied one route to the data, the agent kept finding another:

1. **The agent read the leaked connection credentials and hit BC's OData `/api` directly** from a
   shell. Fix: `agent_subprocess_env()` scrubs every `BC_SERVER_*` / `BC_MCP_*` / `BC_CONTAINER_NAME`
   from the launched agent's environment, and a **credential-free localhost MCP gateway**
   (`src/bcbench/agent/shared/mcp_gateway.py`) fronts BC — it injects the auth headers upstream (so the
   agent's MCP config carries no secret, and nothing is recoverable from the process command line) and
   path-restricts to `/mcp` (so `/api` is unreachable through it).
2. **The agent then ran `docker exec … sqlcmd` against the container's SQL database.** Mitigated by
   withholding the container name and removing the incentive (a working, easy MCP path); the durable
   fix is network isolation, documented as the end-state.
3. **On the Copilot CLI, custom MCP servers wouldn't load at all** — the CLI's MCP registry policy
   fetch returns `403` for the Actions `GITHUB_TOKEN`, blocking every custom server
   ([copilot-cli#4346](https://github.com/github/copilot-cli/issues/4346)). Fix: feed a Copilot-licensed
[...]

PR #856, opened 2026-09-04
Add a modified-only dispatch input to the PR review evaluation

gggdttt:
## Summary

Adds an opt-in `modified-only` dispatch input to `pr-review-evaluation.yml` and wires it to the
`get-entries` reusable workflow, which already implements `--modified-only`.

This is the entry-selection half of the closed PR #849, resubmitted on its own. The other half of
#849 — an `engine-ref` input that overrode the pinned engine at run time — is deliberately **not**
included and should stay out: a run-time override lets the recorded pin and the executed pin
diverge, which is the exact property the BC-Bench -> BC-ALAgents -> BCQuality chain exists to
protect, and which PR #851 is hardening by exporting the pinned identity. Moving the engine pin
remains a pull request that edits `install-agent-harnesses/action.yml`.

Entry selection carries no such hazard. `bcbench dataset list --modified-only` already exists
(`src/bcbench/commands/dataset.py` L31-52) and `get-entries.yml` already exposes the flag
(L8, L49, L63-64); only `pr-review-evaluation.yml` never passed it through, so the code-review
category had no cheap way to evaluate just the entries a pull request touches.

Motivation: the offline eval gate for self-improvement pull requests (ADO 648887, FP3/FP4) needs to
score the handful of gold answers a candidate adds, not all 144 entries.

## Changes

- `modified-only` dispatch input, `default: false`. Existing dispatches are byte-for-byte unaffected.
- Passed through to `get-entries`, where it takes precedence over `test-run` (existing behaviour of
  that workflow, unchanged).
- Added to the `requeue` `workflow-inputs` JSON. **Without this, a `modified-only` run with
  `repeat > 1` would silently re-run the full corpus on every requeued pass**, because requeue
  rebuilds the dispatch from that JSON and would have dropped the flag.
- Concurrency group gains a `modified` bucket. A gate run sets `test-run: false`, so it would
  otherwise land in the `full` group behind a full-corpus run with `cancel-in-progress: false` —
  a gate that queues behind a 144-entry run is not usable as a gate. Runs that do not set
  `modified-only` keep their previous group exactly (`test` or `full`).

## Notes

- **No version bump.** The input is opt-in and default-false, so no existing run's result changes.
  This matches the convention for workflow-only changes (#807, #826, #740, #814, #788, #789);
  `CONTRIBUTING.md` reserves a minor bump for tooling updates that may affect results.
- **No conflict with #851.** That PR edits the `evaluate-with-pr-review` `outputs:` block and the
  `summarize-results` `with:` block; this PR edits the dispatch inputs, the concurrency group, the
  `get-entries` `with:` block, and the requeue JSON. The hunks are disjoint.

## Validation

- `uv run ruff format` / `uv run ruff check` on the touched test file — clean.
- `uv run ty check . --ignore=unresolved-import --exclude "notebooks/"` — 1 diagnostic, pre-existing
  on `main` (`src/bcbench/redteam.py:150`, unused `ty: ignore`); unrelated to this change.
- `uv run pytest -q -m "not e2e"` — 875 passed, 2 skipped, 1 deselected.
- `yaml.safe_load` of the workflow asserts the file still parses and confirms the resolved input
  list, concurrency expression, and `get-entries` inputs.
- New test `test_pr_review_workflow_propagates_modified_only` pins both wiring points, including the
  requeue JSON, so the full-corpus regression above cannot silently return.
- `git diff --stat` — 2 files, 18 insertions, 3 deletions; no formatter noise.

### DIFF
--- a/.github/workflows/pr-review-evaluation.yml
+++ b/.github/workflows/pr-review-evaluation.yml
@@ -25,6 +25,11 @@ on:
         required: false
         default: true
         type: boolean
+      modified-only:
+        description: "Only evaluate entries added or modified against origin/main (takes precedence over test-run)"
+        required: false
+        default: false
+        type: boolean
       repeat:
         description: "Number of times to run sequentially (ignored for test runs)"
         required: false
@@ -43,7 +48,7 @@ on:
         type: string
 
 concurrency:
-  group: pr-review-evaluation-${{ inputs.test-run && 'test' || 'full' }}
+  group: pr-review-evaluation-${{ inputs.modified-only && 'modified' || inputs.test-run && 'test' || 'full' }}
   cancel-in-progress: false
 
 env:
@@ -62,6 +67,7 @@ jobs:
   get-entries:
     uses: $/.github/workflows/get-entries.yml
     with:
+      modified-only: ${{ inputs.modified-only }}
       test-run: ${{ inputs.test-run }}
       category: code-review
 
@@ -162,4 +168,4 @@ jobs:
       repeat: ${{ inputs.repeat }}
       existing-tag: ${{ needs.pin-commit.outputs.tag-name }}
       workflow-inputs: |
-        {"model": "${{ inputs.model }}", "test-run": "${{ inputs.test-run }}", "git-ref": "${{ inputs.git-ref || github.ref_name }}"}
+        {"model": "${{ inputs.model }}", "test-run": "${{ inputs.test-run }}", "modified-only": "${{ inputs.modified-only }}", "git-ref": "${{ inputs.git-ref || github.ref_name }}"}
--- a/.github/workflows/pr-review-evaluation.yml
+++ b/.github/workflows/pr-review-evaluation.yml
@@ -26,7 +26,7 @@ on:
         default: true
         type: boolean
       modified-only:
-        description: "Only evaluate entries added or modified against origin/main (takes precedence over test-run)"
+        description: "Only evaluate entries added or modified against origin/main (takes precedence over test-run). Scores a partial corpus, so the run never publishes results or requeues."
         required: false
         default: false
         type: boolean
@@ -61,7 +61,8 @@ jobs:
       contents: write
     with:
       agent: pr-review
-      test-run: ${{ inputs.test-run }}
+      # A modified-only run never requeues, so it must not leave an ephemeral tag behind.
+      test-run: ${{ inputs.test-run || inputs.modified-only }}
       repeat: ${{ inputs.repeat }}
 
   get-entries:
@@ -139,7 +140,7 @@ jobs:
         with:
           name: evaluation-results-${{ github.run_id }}-${{ matrix.entry }}
           path: ${{ env.EVALUATION_RESULTS_DIR }}/**/*.jsonl
-          retention-days: ${{ inputs.test-run && 1 || 30 }}
+          retention-days: ${{ (inputs.test-run || inputs.modified-only) && 1 || 30 }}
 
   summarize-results:
     needs: evaluate-with-pr-review
@@ -151,14 +152,16 @@ jobs:
       results-dir: ${{ needs.evaluate-with-pr-review.outputs.results-dir }}
       model: ${{ inputs.model }}
       agent: "BC PR Review"
-      mock: ${{ inputs.test-run }}
+      # Scores only the modified entries, so publishing it to Braintrust/Kusto or the
+      # leaderboard would record a partial corpus as a benchmark result.
+      mock: ${{ inputs.test-run || inputs.modified-only }}
       category: code-review
       git-ref: ${{ inputs.git-ref || github.ref_name }}
     secrets: inherit
 
   requeue:
     needs: [summarize-results, pin-commit]
-    if: ${{ !cancelled() && !failure() && !inputs.test-run }}
+    if: ${{ !cancelled() && !failure() && !inputs.test-run && !inputs.modified-only }}
     uses: $/.github/workflows/requeue-evaluation.yml
     permissions:
       contents: write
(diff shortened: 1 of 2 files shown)


.github/workflows/pr-review-evaluation.yml
```diff
@@ -162,4 +168,4 @@ jobs:
       repeat: ${{ inputs.repeat }}
       existing-tag: ${{ needs.pin-commit.outputs.tag-name }}
       workflow-inputs: |
-        {"model": "${{ inputs.model }}", "test-run": "${{ inputs.test-run }}", "git-ref": "${{ inputs.git-ref || github.ref_name }}"}
+        {"model": "${{ inputs.model }}", "test-run": "${{ inputs.test-run }}", "modified-only": "${{ inputs.modified-only }}", "git-ref": "${{ inputs.git-ref || github.ref_name }}"}
```
haoranpb: Curious what is the use case here for repeated runs with modified entries.

The original design for modified entries was so that we could verify the shape of those newly added tasks in the dataset, e.g. the pinned commit is available on remote.

Once we verify that (1 run is usually enough), then we will merge the code so we could trigger full repeated runs.

haoranpb (changes requested):
Search for all places where we use `input.test-run`, because a modified only run is probably also a test job.

`pin-commit` should also consume `inputs.modified-only`, otherwise it will leave behind an ephemeral tag

Similarly, `summarize-results`

.github/workflows/pr-review-evaluation.yml
gggdttt: You're right, there is no use case — I added the key defensively rather than by design, and that was the wrong call.

Your framing is the correct one: one run verifies the shape of the new entries, then they merge and repeats run against the full corpus. I've made that explicit in the workflow instead of leaving it implicit, so `requeue` is now skipped for a modified-only dispatch:

```yaml
if: ${{ !cancelled() && !failure() && !inputs.test-run && !inputs.modified-only }}
```

With requeue unreachable, the payload key was dead code, so it's removed and this line is back to its original form. A test pins the guard, which is what protects against someone relaxing it later and silently re-running all 144 entries:

```python
assert "!inputs.test-run && !inputs.modified-only" in workflow
assert "modified-only" not in workflow.split("workflow-inputs:")[1]
```

`631f3d9d`

gggdttt:
Thanks — the sweep found a worse problem than the leftover tag, and you were right that a modified-only run is essentially a test job.

## `summarize-results` (the serious one)

`mock` gates three things in `summarize-results.yml`, not just artifact retention:

- `--use-capi ${{ !inputs.mock && '--storage braintrust --storage kusto' || '' }}`
- `Update leaderboard in a new branch`, `if: ${{ !inputs.mock && !inputs.skip-leaderboard }}`
- the `evaluation-summary` retention days

With `mock: ${{ inputs.test-run }}` a modified-only run had `mock=false`, so a two-entry run would have been published to Braintrust and Kusto and committed to `docs/_data/code-review.json` as a benchmark result for that model. That is corruption of the record, not just noise. Now:

```yaml
mock: ${{ inputs.test-run || inputs.modified-only }}
```

The `evaluation-summary` artifact is still uploaded under `mock`, so consumers that read the summary in-run are unaffected.

## `pin-commit` and `requeue`

Rather than only suppressing the tag, I disabled the requeue path outright — per your other comment, one run is the point of a modified-only dispatch:

```yaml
requeue:
  if: ${{ !cancelled() && !failure() && !inputs.test-run && !inputs.modified-only }}

pin-commit:
  test-run: ${{ inputs.test-run || inputs.modified-only }}
```

These two go together: with requeue skipped, nothing would ever reach `cleanup-ephemeral-tag`, so `pin-commit` must not create the tag in the first place. That also removes the `exit 1` ("No ephemeral tag is available for requeue") that suppressing the tag alone would have caused on `repeat > 1`.

## Rest of the sweep

The remaining `inputs.test-run` sites in this workflow:

- `retention-days` on the evaluation results artifact — changed, a partial run is short-lived.
- `mock` — changed, above.
- `get-entries` `test-run:` — left alone; `get-entries.yml` already gives `modified-only` precedence.
- concurrency group — already a separate `modified` group in the first commit, so a gate run does not queue behind a full-corpus run.

`dataset-validation.yml` is the existing precedent and needs nothing: it uses `modified-only` with no summarize or publish step at all, which is the same "verify, don't publish" shape this now has.

The input description records the semantics, and `test_pr_review_workflow_treats_modified_only_as_a_partial_run` pins all four wiring points.

`120d082a`, `631f3d9d`. 876 passed, 2 skipped.

.github/workflows/pr-review-evaluation.yml
```diff
@@ -25,6 +25,11 @@ on:
         required: false
         default: true
         type: boolean
+      modified-only:
+        description: "Only evaluate entries added or modified against origin/main (takes precedence over test-run). Scores a partial corpus, so the run never publishes results or requeues."
```
haoranpb: [nit] this looks a bit wired when description is long

haoranpb (dismissed):
Good addition!

.github/workflows/pr-review-evaluation.yml
gggdttt: Agreed, thanks for the screenshot. Shortened to `Only evaluate added or modified entries (never publishes results)` in e0eda507, which now fits the form like the neighbouring inputs and is close to the wording `dataset-validation.yml` already uses.

I kept the non-publishing half because it is the one thing a caller cannot infer from the label, and dropped the rest: the precedence over `test-run` and the requeue behaviour are already recorded as comments next to the wiring they explain, which is where someone changing that behaviour will actually be looking.

haoranpb (approved):

PR #858, opened 2026-09-07
Refresh production PR Review leaderboards

dayland:
Publishes the complete BC PR Review v0.10.0 baseline refresh from immutable BC-Bench commit `06a8bf298db822e27c2be3970f5e9c7a66869f4d`.

- Adds three production runs for each of the eight curated models
- Replaces the earlier partial Luna and Gemini baseline
- Preserves all 80 direct-agent historical runs and 25 experiment runs
- Retains the exact Copilot CLI, BC-ALAgents, and BCQuality identity on every refreshed run and aggregate
- Leaves advanced diagnostics unavailable where the original run artifacts did not preserve them

This PR is independent of #851 and can merge directly to `main`. The retained harness identity fields will become visible in the dashboards when #851 merges.


haoranpb (approved):
This PR is not 100% independent of https://github.com/microsoft/BC-Bench/pull/851 , specifically the added fields 

> "benchmark_commit": null,
> "copilot_cli_version": null,
> "bc_alagents_repository": null,
> ...

but should be harmless to have them

dayland:
> This PR is not 100% independent of #851 , specifically the added fields
> 
> > "benchmark_commit": null,
> > "copilot_cli_version": null,
> > "bc_alagents_repository": null,
> > ...
> 
> but should be harmless to have them

Agreed, I left the metrics but they are benign until/if we enable the added UX

PR #860, opened 2026-09-07
Track harness versions independently of experiment configuration

haoranpb:
## Summary

Record a generic `agent_version` independently of `ExperimentConfiguration`, so agent harness (e.g. BC PR Review engine) can be compared against the same benchmark without aggregating different versions together.

## Why

PR Review engine is a special harness comparing to Copilot CLI, the experimentation happens in a difference repo and we only gets a SHA in BC-Bench, so `ExperimentConfiguration` is always `None`.

It is unrealistic to require the team to always bump the benchmark version when doing an experiment, instead, we should offer experimenting with harnesses' version.

## Extra convenience

Noticed that team was considering an easier way to experiment with a private branch with PR Review Agent in #849 , we provide a convenient way to do that now as input to the workflow.

Note: providing the SHA as the workflow input is designed for convenience only, because no description is provided for those runs. To experiment with persisted result, follow the [Experiment](https://github.com/microsoft/BC-Bench/blob/main/EXPERIMENT.md)


.github/actions/install-agent-harnesses/action.yml
```diff
@@ -9,6 +15,17 @@ outputs:
 runs:
   using: composite
   steps:
+    - name: Validate BC-ALAgents revision
+      id: engine-sha
+      shell: pwsh
+      env:
+        ENGINE_SHA: ${{ inputs.engine-sha || 'ecf8e31759d6ddd6d78e3a0b7836b40134368009' }}
```
gggdttt, line 22: Could we pull this SHA into a named constant like `DEFAULT_ENGINE_SHA` instead of putting it inline in the expression? It would make the default pin easier to spot and update.

src/bcbench/results/summary.py
```diff
@@ -91,6 +92,7 @@ def _base_fields(cls, results: Sequence[BaseEvaluationResult], run_id: str) -> d
             "category": first_result.category,
             "model": first_result.model,
             "agent_name": first_result.agent_name,
+            "agent_version": first_result.agent_version,
```
gggdttt, line 95: Could we check that all results have the same agent version before summarizing? If mixed-version artifacts get collected, this labels the combined scores with just the first version, and the later grouping can no longer separate them. Mixing a known version with `None` should fail too.

src/bcbench/results/leaderboard.py
```diff
@@ -26,6 +26,7 @@ class LeaderboardAggregate(BaseModel, ABC):
     model: str
     agent_name: str
     category: EvaluationCategory
+    agent_version: str | None = None
```
gggdttt, line 29: Could we also show the agent version in the leaderboard tables? They still only display the benchmark version, so these newly separated groups can appear as identical labels with different scores. A short SHA or CLI version, with Unrecorded for older results, would make the comparison clear.

gggdttt (approved):

src/bcbench/results/leaderboard.py
haoranpb: I decided to leave this to you to decide how to best display those results, because it is impossible to back fill those data now..

Maybe you'd like to clean up or showcasing your results in other way

.github/actions/install-agent-harnesses/action.yml
haoranpb: I will leave it as it is for now, by leveraging job output, we managed to only have the SHA in exactly one place, so it should be easier than before

src/bcbench/results/summary.py
haoranpb: This is a valid concern but an existing pattern, there shouldn't be a situation where we will be aggregating results with different agent version.

If that were to happen, please let me know so we could investigate

dayland (approved):

PR #862, opened 2026-09-07
Experiment: add area playbooks to fix-bug agent

alexei-dobriansky:
## Stacked dependency

Depends on #845. This PR is based on `private/ventselartur/BugFixingSkillToBCBench`; review only the commits above that branch.

## Experiment

Measures whether area-specific playbooks improve the existing `fix-bug` custom agent on the same 52-task dataset.

- Vendors exact Warehouse, Manufacturing, Costing, and Subscription Billing playbooks from microsoft/BCAppsBugFix#356
- Adds `discover` mode, where the agent loads at most one playbook after locating a matching source path
- Adds `selected` mode, where BC-Bench selects a playbook from an unambiguous `metadata.area` value before investigation
- Leaves broad `inventory` metadata unmapped in selected mode because those tasks span several domains
- Records playbook mode, content revision, and selected playbook ID in evaluation results
- Configures this branch for the production-representative `discover` arm

Direct discovery coverage is 15 of 52 tasks: Warehouse/Tracking 8, Manufacturing 5, Costing 1, and Subscription Billing 1. The dataset, model, tools, bug-fix evaluation pipeline, and scoring remain unchanged from #845.

## Hypothesis

Area guidance should improve resolution on directly covered tasks without affecting uncovered tasks. Selected mode is retained as an upper-bound diagnostic if discovery routing underperforms.

## Validation

- Exact playbook content matches microsoft/BCAppsBugFix#356 at `1cc59c631e58dca7997f93a136826733821d8c00`
- `uv run pytest -q` — 918 passed, 1 skipped, 1 deselected

### DIFF
--- a/src/bcbench/agent/claude/agent.py
+++ b/src/bcbench/agent/claude/agent.py
@@ -6,9 +6,11 @@ import yaml
 
 from bcbench.agent.claude.metrics import parse_stream_output
 from bcbench.agent.shared import (
+    PlaybookUsageTracker,
     agent_subprocess_env,
     build_al_lsp_plugin,
     build_mcp_config,
+    build_playbook_mcp_server,
     build_prompt,
     resolve_config_plugins,
     start_bc_mcp_gateway,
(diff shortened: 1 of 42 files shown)


alexei-dobriansky:
Follow-up from workflow run 34165334991: commit 0b3d8aa prevents the fix-bug agent from delegating/backgrounding unattended work and persists empty AL diffs as failed benchmark results instead of failing the matrix job. Validation: 921 passed, 1 skipped, 1 deselected; all pre-commit hooks passed.

PR #872, opened 2026-09-10
Fix semantic-first code-review matching

gggdttt:
## Summary

Judge every normalized same-file expected/generated and ignored/generated candidate before irreversible one-to-one assignment. A nearer or tied unrelated finding can no longer prevent the correct candidate from reaching the semantic judge.

Only judge-confirmed edges enter the joint assignment. Its objectives are lexicographic: maximize expected matches, then ignored neutralization, then minimize total line distance. Identity-based row/column mapping keeps equal-value comments distinct; unmatched dummy columns allow sparse graphs without selecting forbidden edges. `CodeReviewResult.create` requires explicit `matched_pairs`, `ignored_comments`, and `ignored_matched_pairs`; result construction has no structural-only fallback. Mock evaluation and test fixtures provide explicit known pairs. Obsolete `match_comments` and `unmatched_generated` compatibility helpers are removed.

For `N` gold and `M` generated vertices, let `D = max_edge_distance * min(N, M) + 1`. Ignored matches receive reward `D`, and expected matches receive `(N + 1) * D`, so each higher-priority objective dominates all lower-priority gains.

BC-Bench is bumped from **0.11.0 to 0.12.0** in `pyproject.toml` and `uv.lock`, following the minor-version requirement for evaluation-result changes. No dependency or harness pins changed, and no tag or release was published.

## Deterministic regression evidence

Both reproductions use **fixed/mock semantic verdicts**, separate from the live-judge experiment below:

- `synthetic__style-018` shape: unrelated missing Caption precedes the correct missing ToolTip at line 23.
- `synthetic__error-handling-silent-skip-01` shape: nearer permissions at line 14 competes with the correct migration skip at line 28 for expected lines 17–19.

Reconnecting main's original location-first matcher in memory reproduces `(matched, missed, incorrect) = (0, 1, 2)` for both; the corrected pipeline produces `(1, 0, 1)`. Tests also cover reversed order, crossed semantic pairs, multiple files, duplicate-value verdict mapping, expected/ignored competition, empty buckets, and judge failure propagation. An exhaustive comparison checks all 512 small candidate graphs against the brute-force three-objective optimum.

## Validation

- Latest revision `671b7a36a63df14562936bf1df0890d04783c205`: focused scorer, calibration, result factory/summary/serialization/hierarchy/writer, PR-review reporting, mock evaluation and version suites: **328 passed, 1 skipped**. The opt-in live calibration test remains disabled; the separate live-judge experiment below was run explicitly.
- All applicable changed-file pre-commit hooks passed, including Ruff, lock consistency and repository type checking.
- The initial scorer implementation received an independent read-only code review with no significant issues.
- After the review-requested API cleanup, all 145 sealed candidate verdicts were replayed offline on `671b7a36`. Every final pair, score tuple and constructor category metric remains identical to the measured `d6b349e4` implementation: **94 / 234 matched**. This replay made **0 new model calls**; all 1,823 sealed experiment files retained their hashes. The original live experiment remains attributed to `d6b349e4`, not relabeled as a new evaluation of the latest head.

## Live-judge frozen-output A/B (2026-09-10)

One complete paired round reused all **145 entries, 234 expected findings and 272 already-generated comments** from [run 34335052023](https://github.com/microsoft/BC-Bench/actions/runs/34335052023). No review-agent regeneration, gold edits, severity changes or manual adjudication labels were used.

- Baseline: `79f36146192a67964345ccfa12750324a917e511`.
- Candidate measured in this live experiment: `d6b349e4fc2c82296e030cfe07d7626ef456aeb4`.
- Actual judge: `gpt-5.3-codex`, identical production prompt and parser on both arms.
- Controlled profile: empty isolated per-arm directories, only the verified `apply_patch` result-file writer available; source reads, custom instructions, MCP/remote and other tools disabled. This is a comments-only scorer experiment, not a full replay of the production review workspace.
- 182 actual judge subprocesses (91 per arm), alternating arm order, maximum 3 concurrent, **0 failures and 0 retries**. Four earlier setup/pilot calls were excluded. All 145 entries are included; no failed rows or unfavorable verdicts were dropped.

| Metric | Baseline | Candidate |
| --- | ---: | ---: |
| Matched / expected | 82 / 234 | 94 / 234 |
| Recall | 35.04% | **40.17%** |
| Precision | 30.15% | **34.56%** |
| F1 | 32.41% | **37.15%** |

This recovers **12 scoring credits across 9 entries, with no entry-level losses**: recall **+5.13 percentage points**. Reusing the candidate arm's actual live verdicts with the old assignment also yields **82 / 234**; all 141 common selected-edge verdicts agree across arms. This shared-verdict control supports attribution to assignment rather than merely a favorable judge rerun. The historical 84 / 234 is only a reference, not this experiment's A/B baseline.

All 12 newly credited pairs were inspected: 10 are direct same-issue matches, one describes the same credential flow with a broader anchor, and one describes the same unnecessary record reload but emphasizes stale-record correctness rather than performance. The latter two semantic boundaries remain explicit; these measurements do not establish that every gold rule or generated explanation is correct.

## Scope and tradeoff

Candidate count is deliberately `sum_f((expected_f + ignored_f) * generated_f)`, without top-k or distance pruning. In this controlled round:

| Judge-only cost | Baseline | Candidate |
| --- | ---: | ---: |
| Observed AI credits | 137.08 | 154.08 (+12.4%) |
| Total model tokens | 839,683 | 903,446 (+7.6%) |
| Mean elapsed time per invoked judge | 19.18 s | 21.53 s |

The paired run took 22.6 minutes at maximum concurrency 3. These are judge-only measurements for this dataset/profile, not full reviewer costs or production latency guarantees; larger candidate sets still carry context-capacity risk.

This is **one paired round measuring recovered scoring credit on fixed outputs**, not improved reviewer generation, repeated-run statistical evidence or an official leaderboard update. Raw prompts, live responses, verdict/index maps, per-entry deltas, hashes and usage records are retained in the originating Copilot session's `live-scorer-ab-34335052023-writer-only` artifacts.

No dataset/gold, knowledge, engine pins, minimum severity, or historical leaderboard/result changes. PR 871's parser behavior is preserved from current main. This is only the first scorer fix under the umbrella bug; gold/knowledge follow-up remains separate.

Related AB#649781: https://dynamicssmb2.visualstudio.com/Dynamics%20SMB/_workitems/edit/649781

This PR does not close the umbrella bug.


src/bcbench/results/codereview.py
```diff
@@ -74,45 +74,57 @@ def _line_distance(line: int, start: int, end: int | None) -> int:
...
-    sentinel = float(finite.max() + 1.0) if finite.size else 1.0
-    solvable = np.where(np.isfinite(cost), cost, sentinel)
-
-    row_indices, column_indices = linear_sum_assignment(solvable)
-    return [
-        (expected_comments[expected_index], generated_comments[generated_index])
-        for expected_index, generated_index in zip(row_indices, column_indices, strict=False)
-        if np.isfinite(cost[expected_index, generated_index])
-    ]
+def match_comments(
```
haoranpb: We don't keep backwards compatbilities, we should be able to safely update all the usages with the new approach

haoranpb (changes requested):
Good Improvement

---

This PR introduces changes that affects evaluation result, we should do at least a minor version bump: https://github.com/microsoft/BC-Bench/blob/main/CONTRIBUTING.md#versioning-policy

src/bcbench/results/codereview.py
gggdttt: Removed `match_comments` and `unmatched_generated` in 671b7a36. `CodeReviewResult.create` now requires explicit `matched_pairs`, `ignored_comments`, and `ignored_matched_pairs`; it no longer infers structural matches. All callers, including mock evaluation and test fixtures, have been migrated. Replaying all 145 sealed candidate verdicts preserves every match and score (94/234), without new model calls.

gggdttt:
@haoranpb Bumped BC-Bench from 0.11.0 to 0.12.0 in pyproject.toml and uv.lock in commit 671b7a36, following the minor-version policy for evaluation-result changes. No dependency or harness pins changed, and no tag or release was published. The compatibility helpers and implicit structural fallback are also removed, with all callers migrated; I replied in and resolved that inline thread. The PR description now reflects the cleanup while keeping the live A/B evidence tied to its original d6b349e4 commit.

haoranpb (approved):

gggdttt:
@haoranpb A versioning question for the follow-up gold corrections under [AB#649781](https://dynamicssmb2.visualstudio.com/Dynamics%20SMB/_workitems/edit/649781).

The scope is three existing fixtures, not a scorer/model change: make `performance-014` actually exercise a scan rather than falsely claiming temporary-table `Get()` is linear; turn `performance-009` into a valid persistent-versus-temporary API boundary instead of requiring all APIs to be temporary; and repair `error-handling-drilldown-position-01` so keys really change across refreshes and its patch is well formed. Existing entry IDs would remain, but one incorrect required finding would become a negative boundary and the other two patches would change.

CONTRIBUTING currently places dataset changes under a major version bump, which would take 0.12.0 to 1.0.0. Earlier gold corrections did not consistently follow that rule. Would you prefer **0.13.0 as an explicitly approved gold-correction exception**, or **1.0.0 per the current policy**? We will not silently change the policy or publish these changed gold cases under 0.12.0. The 0.12.0 scorer release is being kept separate and retains the existing gold.

haoranpb:
@gggdttt I think `0.13.0` would probably be the best call here, we are not ready to be `1.0.0` so the versioning here needs to be handled a bit specially.

gggdttt:
Thanks @haoranpb — applied the approved 0.13.0 version in microsoft/BC-Bench#878 (commit d3510453), with the exception linked in its description. It remains draft while the independent retired-article annotation cleanup in microsoft/BC-Bench#879 unblocks the unchanged-scoring-input 0.12.0 release. No general versioning-policy change or release/tag publication has been made.

PR #885, opened 2026-09-15
Include low-severity findings in PR Review benchmarks

gggdttt:
## Summary

Default the BC PR Review benchmark to `Low` instead of `Medium`, so its review scope includes low-severity gold findings.

Pass the configured floor to **both** `MINIMUM_SEVERITY` and `AGENT_MINIMUM_SEVERITY`. Previously, the adapter set only the latter, leaving the pinned engine's prompt and knowledge-backed findings at its default `Medium` floor. Both gates now honor the benchmark configuration or explicit local override, regardless of inherited severity environment variables.

Update the CLI help and code-review documentation, and cover the default, explicit overrides, ambient environment values, and Low output mapping with regression tests. Bump BC-Bench from `0.13.0` to `0.14.0` because the default review behavior changes.

## Scope

No changes to the dataset, scoring, historical results, or engine/BCQuality pins. Production BC-ALAgents configuration is handled separately. This PR does not claim a measured score improvement.


haoranpb:
@gggdttt do we close this PR now that we experimented with the change?

